PR #7, opened 2019-10-25
make the arch stuff local

Lokathor:


src/arch.rs
```diff
@@ -0,0 +1,114 @@
...
+unsafe impl Pod for Align8<[usize; 2]> {}
+#[cfg(target_pointer_width = "32")]
+unsafe impl Pod for Align16<[usize; 4]> {}
+#[cfg(target_pointer_width = "32")]
+unsafe impl Pod for Align32<[usize; 8]> {}
+//
+#[cfg(target_pointer_width = "64")]
+unsafe impl Pod for Align16<[usize; 2]> {}
+#[cfg(target_pointer_width = "64")]
+unsafe impl Pod for Align32<[usize; 4]> {}
```
MaulingMonkey: Why are these conditional?  What have I missed here..?

Lokathor: the type size times the number of elements has to come out to exactly the alignment to prevent there from being any padding, and the size of a usize changes by pointer width.

src/arch/sse.rs
```diff
@@ -0,0 +1,926 @@
...
+}
+
+impl core::fmt::Display for m128 {
+  /// Display formats in offset order.
+  ///
+  /// All `Formatter` information is passed directly to each individual `f32`
+  /// lane being formatted.
+  fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
+    let a: [f32; 4] = cast(self.0);
+    f.write_str("m128(")?;
```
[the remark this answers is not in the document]
Lokathor: Honestly it's there just because sometimes i like to type `{}` instead of `{:?}`

We could change it later, but we'll merge the PR for now.

PR #11, opened 2019-11-13
Use sse min and max, add clamp function

termhn:


Lokathor:
clamp should unconditionally use min and max, which will select the correct paths on sse vs not-sse. then it can also be moved to the unconditional code section later down the file, which is where we'd prefer as much if the code to live as possible.

termhn:
:+1:

Issue #12, opened 2019-11-14
Test that lane ordering is consistent for `from` and `new` in all cases.


Lokathor:
We probably screwed it up somewhere.

PR #13, opened 2019-11-14
V next

Lokathor:
* Makes wide `sin_cos`, which also allows wide `sin`, `cos`, and `tan`.
* Fixes a bug in `merge` (arguments were swapped from the order stated in the docs)
* Fixes `new` (now argument order matches the indexing order of the created f32x4)
* Introduces an anemic `i32x4` type (just enough to support `sin_cos`)
* Adds a few tests of the `f32x4` type _itself_, not just the `arch` module types.
* `is_nan`, `is_ordinary`

???

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "wide"
 description = "A crate to help you go wide."
-version = "0.2.7-alpha.0"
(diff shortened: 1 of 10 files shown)


Lokathor:
Closes https://github.com/Lokathor/wide/issues/12

PR #16, opened 2019-11-27
Move the crate to use the Zlib license

Lokathor:


Lokathor:
@termhn

Lokathor:
released as 0.3.2

PR #27, opened 2020-04-26
Return value from sin_cos

akikoskinen:
Commit 04355fb963f1 changed the implementation of `sin_cos` but forgot to specify a return value. This commit fixes that.


Lokathor:
So, a favorite streamer of mine has this video that she plays at the start of her stream. and it's like a 30 second quick clip show of stuff she's done in the past. And one of the bits in the clips is her saying "I'm a dumb, I'm such a dumb."

Anyway, no reason why I'd mention that right now.

PR #28, opened 2020-05-03
Fix sin_cos

### LINKED ISSUE #26: Sin and cos give incorrect results?
I tried to use the `wide::sin_f32` and `wide::cos_f32` functions. I got some confusing results so I plotted some of the results. This plot shows the functions' results as well as the real sin and cos values on the interval -6.5..6.5 (about minus 2*pi to 2*pi):

Wide sin and cos plot

What gives?

akikoskinen:
Tracked down the reason why `sin_cos` was giving incorrect results for some angles (#26). The issue was with the `i32x4` left shift function. The amount to be shifted was constructed in a wrong way and as a result it always just zeroed the to-be-shifted value (by shifting the bits way too much, and "out" of 
the left end of the value).

Also added tests for all these related parts.

Fixes #26.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -3,7 +3,6 @@ use wide::*;
 fn main() {
   // /*
   let mut f = 0.0;
-  #[cfg(feature = "extern_crate_std")]
   while f <= 1.5 {
     println!("rad {:.2}: s/c {:?}", f, f32x4::from(f).sin_cos());
     f += 0.1;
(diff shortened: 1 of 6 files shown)


Lokathor:
Ohhhhhhhh, I completely misunderstood how the intrinsic worked. I thought that each lane was supposed to be its own shift amount, but it reads a single shift amount from the register as an i64 (and shifts of more than 32 are useless do i don't know why it reads more than the bottom few bits).

Actually I've been going over the intrinsics again and making free functions versions of each with docs and so forth [in another repo](https://github.com/Lokathor/safe_arch), and the inner `arch` module will get replaced with that repo once it's complete enough. If i'd done that to begin with and not cut corners by skipping doctests we could have avoided this.

There's a trillion intrinsics but even then don't cut corners kids.

MANY many thanks.

Lokathor:
`0.4.6` is now out.

PR #39, opened 2020-08-19
Added exp() function to f32x4 and f64x2 inspired/copied from Agner Fo…

ronniec95:
…g vcl library with corresponding tests. Ready for review.

### DIFF
--- a/src/f32x4_.rs
+++ b/src/f32x4_.rs
@@ -27,6 +27,16 @@ macro_rules! polynomial_2 {
   }};
 }
 
+macro_rules! polynomial_5 {
+  ($x:expr, $c0:expr, $c1:expr, $c2:expr, $c3:expr, $c4:expr, $c5:expr $(,)?) => {{
+    let x = $x;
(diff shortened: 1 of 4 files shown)


Lokathor:
This is good. I might poke a few spots around just for style, but i can do that myself later. Should I merge it now?

ronniec95:
Go for it. I'm use **rustfmt** for styling incidentally - do you use something else?

Lokathor:
oh the rustfmt is fine, but I'll probably move the polynomial functions up to lib.rs now that there's some crossover between f32x4 and f64x2.

Lokathor:
Also, I can't publish a crate update until later today (~6 hrs) so if you want to get the ln/log stuff into the same update there's time available.

ronniec95:
I've done the log functions but it's not yet tested. It's late here too so I'll have to pick it up tomorrow if that's OK. At 0.5.2 release for the ln functions is fine :)

Lokathor:
okay, no rush then

PR #45, opened 2020-08-20
Ln/log10/log2 added to f32x4 and f64x2

ronniec95:
Added natural log (ln(e)) log10 log2 for f32x4 and f64x2
Added polynomial5n and 8 to lib.rs


src/f32x4_.rs
```diff
@@ -622,4 +622,74 @@ impl f32x4 {
...
+
+  #[inline]
+  #[allow(non_upper_case_globals)]
+  fn exponent(self) -> f32x4 {
+    let t1 = cast::<_, u32x4>(self);
+    let t2 = t1 << 1;
+    let t3 = t2 >> 24;
+    // There might be a simpler way to do this cast from i32 back to f32 using bytemuck?
+    let t4: [i32; 4] = cast(cast::<_, i32x4>(t3) - i32x4::from(0x7F));
+    f32x4::from([t4[0] as f32, t4[1] as f32, t4[2] as f32, t4[3] as f32])
```
Lokathor, line 634: Here do you want to preserve the numeric value as you go from i32 to f32? I believe there's a `round_float` method on the i32x4 type which does what you want.

tests/t_f32x4.rs
```diff
@@ -396,3 +396,23 @@ fn test_f32x4_none() {
...
+    (1.0, 0.0),
+    (2.718282, 1.0),
+    (10.0, 2.3025851),
+    (35.0, 3.5553482),
+    (1250.0, 7.130899),
+  ]
+  .iter()
+  .copied()
+  {
+    let expected = f32x4::from(e);
```
Lokathor, line 414: I think for the tests we should compare for a minimal difference compared to the stdlib version of the same function. Have a look at how I reworked the `exp` test for an example, though the difference threshold might be different in this case.

src/f32x4_.rs
ronniec95: Thanks - fixed and pushed

tests/t_f32x4.rs
ronniec95: Great idea - done and pushed

ronniec95:
It's pretty accurate compared the std.rs. 1e-11 diff on f64 and 1e-6 on f32. I've tested the performance of this on a blackscholes option pricer (which I'll release in due coarse) and the timing went down from 385ms for 16m calculations to 225ms which is good while maintaining the accuracy.

Happy for you to merge.

Issue #43, opened 2020-08-19
Boolean masks should become their own type


Lokathor:
some people like the increased type safety.

this is a breaking change so it'd be in 0.6, but otherwise it's not hard.

Lokathor:
Note: This will break because all existing `cmp_eq` operations would have to change from returning Self to returning a mask value.

Lokathor:
Probably we can just have the "mask" factor be a wrapper type over a SIMD type, with blend / bit functions.

```rust
pub struct Mask<T>{ inner: T }
```

Lokathor:
I don't think this will end up happening any time soon, realistically.

PR #48, opened 2020-08-21
Implement {float}::asin_acos(), {float}::asin(), and {float}::acos()

### LINKED ISSUE #44: {floating}::acos()
Needed for spherical linear interpolation

termhn:
Fixes #44


termhn:
@Lokathor i rebased this on the avx changes and added impls for those types as well, and I also #[cfg]'d out the tests that don't pass for now with FIXME's to put them back in once #43 lands

Lokathor:
whoops forgot about this.

we good here?

termhn:
Believe so :)

PR #51, opened 2020-08-30
Reduce add

ronniec95:
Horizontal Add functionality which is 2x-4x faster than iter.sum() in my benches


ronniec95:
Hey Lokathor

Feel free to merge this and my other PR. Both should now be working a bit more robustly.

Horizontal/reduce add seems to be faster than casting to an array and summing. It's particularly seems faster when doing an iter.sum() on an array of floats/doubles often by 4x. Maybe that's something to take back to `rustc` team as I don't think we should be able to write better code by hand

PR #52, opened 2020-09-01
Minor fixes

ronniec95:
For consistency

Moved polynomial_4 to lib.rs so it can be used with f64x* later
Made sign_bit public which was an oversight on my part


Lokathor:
This is good work.

The one note I think I'd make is that the `i8x32` (and perhaps other types) has alignment that changes based on if `avx2` is available or not. I think that we should use `repr(C, align(x))` to set the alignment to always be the higher value, just to keep it consistent across platforms.

Similarly, `f32x4` has align 16 on all platforms, but actually on ARM the type would only need to be align 8 (assuming that rust had ARM instrinsics in stable and such).

ronniec95:
I'll look at the alignment over the next few days. Let this cook for a bit, I'm adding features as I use them in my main project and I'm finding bugs; so wait a week or so before merging this

Lokathor:
they shouldn't be hard  but it's breaking is all. I can probably do it this weekend.

Lokathor:
In https://github.com/Lokathor/wide/pull/53 you said that this should be merged first. Is this one ready?

ronniec95:
I've been using for 4 days or so against my code and seems ok. Obviously it's not battle tested, but good enough for now.

PR #53, opened 2020-09-07
Ops with base types

ronniec95:
A quick PR to add operations with base types. Implements the base Add/Sub/Mul/Div traits for base types.

Probably need to merge the other PR first if you are happy with it.


ronniec95:
This PR may result in slightly improved performance as you can do examples like
```
    const C: f32 = 0.3989422804014330;
    (f32x8::HALF * e * e).exp() * C
```

I think I would also want to be able to do the f32 * f32x8 operation too so that's a todo... I'll see if I have some time tomorrow

Lokathor:
Yeah I wanted to do this sooner but I wasn't sure if people would be fine with the types being a little more "loose" like this.

Lokathor:
Alright, we ready?

ronniec95:
Yep. Please merge. Thank you

Lokathor:
released as `0.5.4`

PR #67, opened 2020-10-11
Add f32x4::trunc_int and f32x8::trunc_int

### LINKED ISSUE #66: Add trunc method
Specifically `_mm_cvttps_epi32`.

Looks like it missing or I could not find it.

RazrFalcon:
Closes #66

### DIFF
--- a/src/f32x4_.rs
+++ b/src/f32x4_.rs
@@ -518,6 +518,33 @@ impl f32x4 {
       }
     }
   }
+  // Allow only for x86 with SSE and for other architectures when the `std` feature is enabled.
+  // This is done to exclude i586, which retuns invalid results for f32::trunc.
+  #[cfg(any(
+    target_feature = "sse",
+    all(feature = "std", not(any(target_arch = "x86", target_arch = "x86_64")))
+  ))]
+  #[inline]
+  #[must_use]
+  pub fn trunc_int(self) -> i32x4 {
+    pick! {
+      if #[cfg(target_feature="sse2")] {
+        cast(truncate_m128_to_m128i(self.sse))
+      } else {
+        let rounded: [f32;4] = cast(self.round());
+        let rounded_ints: i32x4 = cast([
+          rounded[0].trunc() as i32,
+          rounded[1].trunc() as i32,
+          rounded[2].trunc() as i32,
+          rounded[3].trunc() as i32,
+        ]);
+        cast::<f32x4, i32x4>(self.is_finite()).blend(
+          rounded_ints,
+          i32x4::from(i32::MIN)
+        )
+      }
+    }
+  }
   #[inline]
   #[must_use]
   pub fn mul_add(self, m: Self, a: Self) -> Self {
--- a/src/f32x8_.rs
+++ b/src/f32x8_.rs
@@ -630,6 +630,37 @@ impl f32x8 {
       }
     }
   }
+  // Allow only for x86 with SSE and for other architectures when the `std` feature is enabled.
+  // This is done to exclude i586, which retuns invalid results for f32::trunc.
+  #[cfg(any(
+    target_feature = "avx",
+    all(feature = "std", not(any(target_arch = "x86", target_arch = "x86_64")))
+  ))]
+  #[inline]
+  #[must_use]
+  pub fn trunc_int(self) -> i32x8 {
+    pick! {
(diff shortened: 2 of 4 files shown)


RazrFalcon:
I've added this method only to f32x4, because I have no idea how it should be implemented for other types.

Lokathor:
hmm, since normal `trunc` is f32 output, this should probably be called `trunc_int` to indicate it's changing the type.

funny that you'd submit this today of all days, stdsimd has a similar pr just this morning: https://github.com/rust-lang/stdsimd/pull/34

You should be able to do this at least for f32x8 I think.

And yeah, so you'd have to cfg around this (i forget if `wide` has a `std` feature yet?) or implement a truncation function. You could use `libm` for inspiration here.

RazrFalcon:
>this should probably be called trunc_int

Ok.

Yes, this is an interesting coincidence. I blame the fact that it's weekends. 

Yes, I'm using `std` feature for this method.

I have a strange bug, where `3.7_f32.trunc()` returns 4 on i586. Investigating.

PS: Can I use stdsimd now or is it just for std?

Lokathor:
stdsimd is not yet available in the standard library. You can use it as a stand alone crate on nightly, but i don't think it's published on crates.io even. It's super new and super bare bones. Probably give it a few months to grow.

Lokathor:
and yeah, uh, i586 is problem-town. you could cfg around that if you want, i wouldn't mind. Just make the method not exist at all on that platform (and ditto on the test). if someone actually uses i586 they can fix it then, but i don't think anyone really does much with it.

RazrFalcon:
Ok, I will wait for it.

>Just make the method not exist at all on that platform

Is there a way to do this?

Lokathor:
You can place a cfg on the method itself

```rust
#[crg(cond)]
fn foo(self) -> out {
  //
}
```
then it will just only exist on select targets

RazrFalcon:
Yes, of course. I mean the actual condition. We don't have i586 `target_arch`.

Lokathor:
oh! x86 or x86_64 target_arch but *not* target_feature sse

RazrFalcon:
Done. I've also added it to f32x8.

RazrFalcon:
On the other hand, I actually need a fallback implementation.

RazrFalcon:
It passes all the tests, but it will not compile on ARM...

Lokathor:
actually, sorry, i was answering via my phone for a bit and just sat down and re-read it properly.

Are you saying that the standard library `f32::trunc` can give bad results on `i586`?

RazrFalcon:
Yes, it returns 4 for 3.7

https://github.com/Lokathor/wide/runs/1239019400

Lokathor:
That's uh... a fun bug. @workingjubilee you might add that to your collection of i586 weirdness

RazrFalcon:
Yes... So the question is how to ignore i586 via `cfg`.

Lokathor:
So, on the function you'd probably want to have
```rust
#[cfg(any(
  target_feature = "sse",
  all(feature = "std", not(any(target_arch = "x86", target_arch = "x86_64")))
))]
```
"if we have sse, or if we have the standard library on some other arch"

RazrFalcon:
And I though that I had to add just one line of code... This is why I want to use an existing SIMD crate instead of custom one.

RazrFalcon:
Looks like it's done.

Lokathor:
Released `wide-0.5.6`

workingjubilee:
> this utter nonsense of a bug

LOVIN' IT.

Lokathor:
Correction: `wide-0.5.6` is yanked, released as `wide-0.6.0`

programmerjake:
> Yes, it returns 4 for 3.7
> 
> https://github.com/Lokathor/wide/runs/1239019400
> 
> PS: i586 is tier2, so anything could happen, I guess.

It's obviously wrong since you're `round`ing then calling `trunc`, so `round` converts `3.7` to `4` and then `trunc` converts `4` to `4`. No wonder it causes tests to fail!

RazrFalcon:
@programmerjake Good catch. I've simply copy-pasted the `round_int` code... Still strange that it has different results.

programmerjake:
Also, sorry if I came across as insulting, that wasn't intended

RazrFalcon:
@programmerjake No worry =)

PR #79, opened 2020-10-17
Add SSSE3 fallback to f32x8::trunc_int

RazrFalcon:


Lokathor (approved):
This is good, but note that `ssse3` is *not* on by default in most rust targets, so you'll have to specially enable it if you want this code path to be used.

Lokathor:
Should I merge or do you want to add more elsewhere?

RazrFalcon:
Yes, I'm using `-Ctarget-cpu=native/haswell`.

I guess this patch would be superseded by the next one.

PR #86, opened 2021-06-06
i16x16

### LINKED ISSUE #59: u16x16
This one fits into 256bits, so it can be implemented via AVX.

gilescope:
Doesn't fix #59 but does add i16x16.
(For the test data I've used 32x8's test data twice.)

### DIFF
--- a/src/i16x16_.rs
+++ b/src/i16x16_.rs
@@ -304,14 +304,14 @@ macro_rules! impl_shr_t_for_i16x16 {
               self.arr[5] >> u,
               self.arr[6] >> u,
               self.arr[7] >> u,
-              self.arr[8] << u,
-              self.arr[9] << u,
-              self.arr[10] << u,
-              self.arr[11] << u,
-              self.arr[12] << u,
-              self.arr[13] << u,
-              self.arr[14] << u,
-              self.arr[15] << u,
+              self.arr[8] >> u,
+              self.arr[9] >> u,
+              self.arr[10] >> u,
+              self.arr[11] >> u,
+              self.arr[12] >> u,
+              self.arr[13] >> u,
+              self.arr[14] >> u,
+              self.arr[15] >> u,
             ]}
           }
         }
--- a/src/i16x16_.rs
+++ b/src/i16x16_.rs
@@ -398,8 +398,7 @@ impl CmpLt for i16x16 {
   fn cmp_lt(self, rhs: Self) -> Self::Output {
     pick! {
       if #[cfg(target_feature="avx2")] {
-        //TODO: what about == ?
-        Self { avx2: !cmp_gt_mask_i16_m256i(self.avx2, rhs.avx2) }
+        Self { avx2: !cmp_gt_mask_i16_m256i(self.avx2, rhs.avx2) ^ cmp_eq_mask_i16_m256i(self.avx2,rhs.avx2) }
       } else if #[cfg(target_feature="sse2")] {
         Self { sse0: cmp_lt_mask_i16_m128i(self.sse0, rhs.sse0),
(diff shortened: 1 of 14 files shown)


src/i16x16_.rs
```diff
@@ -0,0 +1,504 @@
...
+  #[inline]
+  #[must_use]
+  fn cmp_lt(self, rhs: Self) -> Self::Output {
+    pick! {
+      if #[cfg(target_feature="avx2")] {
+        //TODO: what about == ?
+        Self { avx2: !cmp_gt_mask_i16_m256i(self.avx2, rhs.avx2) }
+      } else if #[cfg(target_feature="sse2")] {
+        Self { sse0: cmp_lt_mask_i16_m128i(self.sse0, rhs.sse0),
+          sse1: cmp_lt_mask_i16_m128i(self.sse1, rhs.sse1) }
```
gilescope, line 405: am suspicious of this. !gt is not the same as lt. I've done the same as others but wondering if this is a bug?

src/i16x16_.rs
```diff
@@ -0,0 +1,504 @@
...
+
+impl i16x16 {
+  #[inline]
+  #[must_use]
+  pub fn blend(self, t: Self, f: Self) -> Self {
+    pick! {
+      if #[cfg(target_feature="avx2")] {
+        Self { avx2: blend_varying_i8_m256i(f.avx2, t.avx2, self.avx2) }
+      } else if #[cfg(target_feature="sse4.1")] {
+        Self { sse: blend_varying_i8_m128i(f.sse, t.sse, self.sse) }
```
gilescope, line 438: Is this right that blend should be i8?

src/i16x16_.rs
Lokathor: yeah !gt isn't the same as lt, because two i16 can be equal

src/i16x16_.rs
Lokathor: an i8 blend should be fine if the mask is an i16 mask. each two i8 lanes of the mask will be identical, so they'll use the identical source for the output, so it ends up working like an i16 blend.

src/i16x16_.rs
gilescope: Excellent, I've fixed the i32x8 code that I was basing it from. Was a lucky chance that I happened to copy that one as that seemed the only one that wasn't correct.

gilescope:
Does that look ok @Lokathor or would you like u16x16 also? (If you do that's fine though I'd be tempted to put that in a separate PR)

Lokathor:
skipping the u16x16 stuff is fine.

I will try to have a look here later today, but in the meantime you probably are missing a few cfg things since the i586 had trouble in Ci.

src/i16x16_.rs
```diff
@@ -0,0 +1,503 @@
...
+              self.arr[6] >> u,
+              self.arr[7] >> u,
+              self.arr[8] << u,
+              self.arr[9] << u,
+              self.arr[10] << u,
+              self.arr[11] << u,
+              self.arr[12] << u,
+              self.arr[13] << u,
+              self.arr[14] << u,
+              self.arr[15] << u,
```
Lokathor, line 314: 8 to 15 are still shifting left

Lokathor (changes requested):

src/i16x16_.rs
gilescope: ok will add more tests to pick that up.

gilescope: (good spot)

gilescope: Ah I see that the test was there and it was caught by CI. nice.

Lokathor (approved):

PR #97, opened 2021-07-30
Implement Support for WebAssembly SIMD

CryZe:
This adds support for the `+simd128` target feature for WebAssembly. The SIMD intrinsics just got stabilized with Rust `1.54.0`.


CryZe:
Nice, all tests pass. This is ready for review now 🎉

Lokathor (approved):

PR #99, opened 2021-08-01
Activate `std` by default

### LINKED ISSUE #98: Make `std` a default feature
`wide` has two major dependents, `ultraviolet` and `simba`. Neither of those activate `wide`'s `std` feature, despite `ultraviolet` being an `std` only crate, and `simba` having an `std` feature itself. So I'd say, not only is having `std` not be the default a risky move, but in practice, people actually do seem to get it wrong, slowing them down unnecessarily.

CryZe:
It's too risky to have it be deactivated by default as all users of this crate seem to have missed it. This is a breaking change.

Resolves #98

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -58,11 +58,11 @@ jobs:
     - name: Build the crate
       run: cargo build --target ${{ matrix.rust.target }}
 
-    - name: Test with default CPU features.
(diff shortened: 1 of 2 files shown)


Lokathor (approved):

Lokathor:
I think _risky_ is an overstatement, and yet I will merge the PR.

PR #103, opened 2021-08-04
Implement multiplication in most vectors that dont have it

### LINKED ISSUE #102: ``i64x2`` multiplication
this is kind of wasm only (and it translates to [10 instructions](https://github.com/v8/v8/blob/b6520eda5eafc3b007a5641b37136dfc9d92f63d/src/compiler/backend/x64/code-generator-x64.cc#L2834-L2858)) but should this be implemented where possible?

ImUrX:
Here we are, I implemented
* Multiplication for ``i64x2``, ``i64x4``, ``u64x2``, ``u64x4`` ``u32x4`` ``u16x8``
* Splats for all of them so you can simply do ``i64x2 * i64``
* Tests for all the types
----------------------
i think we should implement multiplication in every type that can have it, better done now than later tbh

fixes #102

### DIFF
--- a/src/i16x16_.rs
+++ b/src/i16x16_.rs
@@ -163,6 +163,60 @@ impl Mul for i16x16 {
   }
 }
 
+impl Add<i16> for i16x16 {
+  type Output = Self;
+  #[inline]
+  #[must_use]
+  fn add(self, rhs: i16) -> Self::Output {
+    self.add(Self::splat(rhs))
+  }
+}
+
+impl Sub<i16> for i16x16 {
+  type Output = Self;
+  #[inline]
+  #[must_use]
+  fn sub(self, rhs: i16) -> Self::Output {
+    self.sub(Self::splat(rhs))
+  }
+}
+
+impl Mul<i16> for i16x16 {
+  type Output = Self;
+  #[inline]
+  #[must_use]
+  fn mul(self, rhs: i16) -> Self::Output {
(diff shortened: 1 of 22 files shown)


Lokathor (approved):

ImUrX:
should i continue in implementing Mul to other vectors?

Lokathor:
That would be great.

Also, if you can `impl Mul<simd> for scalar` so that the scalar value can be on the left side of the `*` that would be handy. I think f32x4 has this sort of thing as an example.

ImUrX:
will implement tests later

Lokathor (approved):

ImUrX:
rustfmt returned ``Warning: the `merge_imports` option is deprecated. Use `imports_granularity=Crate` instead``

Lokathor:
Ah, yeah, the file is a little old but i usually have format-on-save set in VS Code so I forget the warning. I'll get it at some point.

ImUrX:
Well it's done, and it doesnt seem like a breaking change :P

Lokathor (approved):

ImUrX:
@Lokathor i dont have write access if you are waiting for that

Lokathor:
Just wasn't sure if you were gonna add more :3

ImUrX:
im gonna check some of the other issues later, but for now, im gonna end what i started in the project i was working on

PR #105, opened 2021-08-05
add slice methods

### LINKED ISSUE #60: as_slice method
Is there a method to get the wide type content as a slice of primitive types? Like `f32x4` -> `[f32; 4]`?

ImUrX:
im thinking of implementing Index and IndexMut but i want to know what you meant about the note to avoid in hot code, by that i mean how to implement that note lol
closes #60

### DIFF
--- a/src/f32x4_.rs
+++ b/src/f32x4_.rs
@@ -1464,7 +1464,7 @@ impl f32x4 {
     Self::pow_f32x4(self, f32x4::splat(y))
   }
 
-  pub fn as_slice(self) -> [f32; 4] {
+  pub fn to_array(self) -> [f32; 4] {
     cast(self)
   }
 }
--- a/src/f32x4_.rs
+++ b/src/f32x4_.rs
@@ -1463,4 +1463,8 @@ impl f32x4 {
   pub fn powf(self, y: f32) -> Self {
     Self::pow_f32x4(self, f32x4::splat(y))
   }
+
+  pub fn as_slice(self) -> [f32; 4] {
+    cast(self)
+  }
 }
--- a/src/f32x8_.rs
+++ b/src/f32x8_.rs
@@ -1681,7 +1681,7 @@ impl f32x8 {
     Self::pow_f32x8(self, f32x8::splat(y))
   }
 
-  pub fn as_slice(self) -> [f32; 8] {
+  pub fn to_array(self) -> [f32; 8] {
     cast(self)
   }
 }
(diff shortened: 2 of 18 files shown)


Lokathor:
I'd kinda forgotten about the details of #60. These are converting the SIMD value to an array. That's useful, but the usual name would in that case be `to_array` instead of `as_slice`. This would be most similar to [to_bits](https://doc.rust-lang.org/std/primitive.f32.html#method.to_bits) in the standard library.

Other than the name the code is all fine.

ImUrX:
yeah its useful, i was thinking ``as_slice`` as the Vec method, but i just noticed it passes a reference instead, so i should really rename it to ``to_array``

i saw that there are instructions for extracting and replacing lanes which could be pretty nice for implementing ``Index`` and ``IndexMut``, should i continue on that idea too?

ImUrX:
But now that i think about it ``Index`` and ``IndexMut`` expect references to the value which I dont think its gonna be possible

Lokathor:
So what happens is that it *works*, but the compiler pushes the register onto the stack, performs the memory operation on the reference, and then pops the value off the stack. It's completely inefficient, which is why `wide` doesn't offer it.

If you still *really* want to do it, a person could use `bytemuck` to fiddle a reference to a SIMD value into a reference to a slice or something like that. But the hope is that the difficulty of doing such a thing will help prevent people from casually using `Index` and then losing a ton of performance.

ImUrX:
you are right, i will just correct the names for this pr

Lokathor (approved):

torokati44:
I found that, in the use case of simply applying a mathematical function to a big array of input numbers, and putting the results into a different big array of output values, the `to_array()` function is really inefficient, as it likely always creates an unnecessary intermediate copy of the lanes (possibly even allocates, even if only on the stack).
And I found no other way of accessing the individual lane values to copy them into the output array. 
But using `bytemuck::cast_ref`, this bulk processing is much faster (on wasm32 target at least), as the values can go straight from input array to vector registers to output array (I haven't verified this though).

Lokathor:
I'm not entirely sure I understand your situation so let me see if I follow what you're saying:
* there's some large slice of data
* you want to iterate the slice in simd chunks and do simd ops
* the results of the simd ops are written to a second slice of data.

In a situation like this, since you already mentioned `bytemuck`, I would suggest using [pod_align_to](https://docs.rs/bytemuck/latest/bytemuck/fn.pod_align_to.html). Then you can do the entire middle portion as an iteration over simd values.

If this doesn't work out for you please open a new issue though. I'm interested in having things work, but old PRs that are already merged is where things quickly get lost and forgotten about.

PR #108, opened 2021-09-13
Improve performance of saturating float to int casts

CryZe:
I randomly stumbled across a Chrome bug ticket where they reported that their lowering of saturating float to int casts is suboptimal:

https://bugs.chromium.org/p/v8/issues/detail?id=12094

They by now merged a fix. The code I ported was based on Chrome and thus the same performance pitfall applies to wide.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,27 +11,27 @@ jobs:
       matrix:
         rust:
         # x86 without sse/sse2 on by default
-        - { target: i586-pc-windows-msvc, toolchain: 1.52.0, os: windows-latest }
+        - { target: i586-pc-windows-msvc, toolchain: 1.52.1, os: windows-latest }
         - { target: i586-pc-windows-msvc, toolchain: stable, os: windows-latest }
         - { target: i586-pc-windows-msvc, toolchain: beta, os: windows-latest }
         - { target: i586-pc-windows-msvc, toolchain: nightly, os: windows-latest }
         # x86
-        - { target: i686-pc-windows-msvc, toolchain: 1.52.0, os: windows-latest }
+        - { target: i686-pc-windows-msvc, toolchain: 1.52.1, os: windows-latest }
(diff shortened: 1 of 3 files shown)


CryZe:
Spurious link error?!

CryZe:
I'm thinking this is the incremental compilation bug in 1.52.0.

CryZe:
Alright, no idea what's wrong then.

CryZe:
Seems like stable is unaffected by whatever weird bug this is in 1.52 though.

Lokathor:
does it build with latest nightly/stable? We can just bump the minimum rust version if it's necessary.

CryZe:
It seems to. Idk if bumping MSRV for a compiler bug makes sense.

CryZe:
Alright, failing now on beta too. I'm out of ideas. Works just fine for me locally.

CryZe:
Maybe the specific native target-cpu uses some instructions or so that the MSVC linker can't handle?!

CryZe:
Googling the error makes it sound like the .exe is still running or something?!

CryZe:
Fun, the problem is unrelated to my changes and seems to happen on the main branch too.

CryZe:
I'm guessing that we end up with the same .exe hash and thus it wants to overwrite the previously created one, but Windows runners have started running Windows Defender or so in the background which locks the file for scanning.

Lokathor:
well, put the changes in and the rest is a cargo issue.

CryZe:
Well more of a Github Actions bug. We may need to use a different target folder for the target-cpu=native, so it doesn't collide.

Lokathor:
Oh. Can we break it up across more CI tasks so that each goes in separate VM instances and they don't clash?

Lokathor:
alternative target folder also works too i guess

CryZe:
Yep, cargo clean fails as well, not being able to remove the executables. So they are definitely being scanned.

CryZe:
This is ready now btw. Using a separate target folder worked.

Lokathor (approved):

PR #109, opened 2021-09-15
Try disabling the Windows CI workaround again

CryZe:
GitHub supposedly turned off some monitoring on their side. This is to test if that indeed fixes their issue.


Lokathor:
Seems green

CryZe:
Yeah, we can either close or merge this.

Issue #111, opened 2021-12-09
Add slice methods? v2


torokati44:
Sorry for necrobumping #105, let's continue here then.

Your understanding of what I'm doing is correct.
I'm processing a planar image with 8 bits per pixel per component of depth, and I plan to speed it up by doing 4 adjacent pixels at a time in SIMD. And since it's a colorspace conversion operation, it needs to duplicate (upscale) chroma samples from the source on load.
If there is a better lane arrangement for this case, please let me know.

Thanks for the `pod_align_to` tip, it's a neat function, and could come in handy, but unfortunately in this case, on each iteration I also have to convert from `&[u8; 4]` to `i32x4`, and then back for storing the results - for more bits of precision to work with in between.
And I doubt that loading an `[u8; 16]`, then throwing out 3/4 of it with a shuffle/interleave would be faster than doing it with a simple element-wise upcast, but that's just my guess.
*EDIT:* Maybe some sort of simple gather operation would also come in handy for this!

And on the store side, I also have to transpose the 3 `i32x4` (4xR, 4xG, 4xB) results to get an interleaved output image.

torokati44:
Basically, writing this:

```rust
let r: &[i32; 4] = bytemuck::cast_ref::<i32x4, [i32; 4]>(&r);
let g: &[i32; 4] = bytemuck::cast_ref::<i32x4, [i32; 4]>(&g);
let b: &[i32; 4] = bytemuck::cast_ref::<i32x4, [i32; 4]>(&b);

// The output alpha values are fixed
rgba.copy_from_slice(&[
    r[0] as u8, g[0] as u8, b[0] as u8, 255, r[1] as u8, g[1] as u8, b[1] as u8, 255,
    r[2] as u8, g[2] as u8, b[2] as u8, 255, r[3] as u8, g[3] as u8, b[3] as u8, 255,
]);
```

instead of this:

```rust
let r: [i32; 4] = r.to_array();
let g: [i32; 4] = g.to_array();
let b: [i32; 4] = b.to_array();

// The output alpha values are fixed
rgba.copy_from_slice(&[
    r[0] as u8, g[0] as u8, b[0] as u8, 255, r[1] as u8, g[1] as u8, b[1] as u8, 255,
    r[2] as u8, g[2] as u8, b[2] as u8, 255, r[3] as u8, g[3] as u8, b[3] as u8, 255,
]);
```

is much faster. It just looks even less neat.
[...]

PR #112, opened 2021-12-12
Add as_array_ref to all types, similar to to_array

torokati44:
As per: https://github.com/Lokathor/wide/issues/111#issuecomment-990446169

I didn't call them `as_slice` because the sizes of all return types are known.


Lokathor:
Oh excellent, this is better than as_slice.

Lokathor:
released in `0.7.2`

torokati44:
Awesome, thanks!

PR #116, opened 2022-10-06
Add inline attribute to array casts

shssoichiro:
I found that the Rust compiler was not inlining the methods `to_array` and `as_array_ref`, which are extremely thin wrappers around cast functions and should in theory always be inlined. The lack of inlining was also preventing certain code optimizations that the compiler would normally perform.


Lokathor:
I added even more `#[inline]` on a lot of other methods too and then released `0.7.5`

PR #118, opened 2022-10-27
Add a faster round fallback when std is available

shssoichiro:


shssoichiro:
> wasm trap: wasm `unreachable` instruction executed

Cursed :skull:

Issue #120, opened 2022-12-28
AVX version of `u32x8::cmp_lt` refers to cmp_eq_mask_i32_m256i


remyoudompheng:
This looks incorrect and makes it return different results from the SSE2 implementation.

```rust
  pub fn cmp_lt(self, rhs: Self) -> Self {
    pick! {
      if #[cfg(target_feature="avx2")] {
        Self { avx2: cmp_eq_mask_i32_m256i(self.avx2, rhs.avx2 ) }
      } else if #[cfg(target_feature="sse2")] {
        Self { sse0: cmp_lt_mask_i32_m128i(self.sse0,rhs.sse0), sse1: cmp_lt_mask_i32_m128i(self.sse1,rhs.sse1), }
```

https://github.com/Lokathor/wide/blob/v0.7.5/src/u32x8_.rs#L350

Lokathor:
Ah, yeah that's definitely incorrect. I'll see if I can fix it soon, or I could accept a PR for it.

Lokathor:
It looks like there aren't SIMD less-than ops for u32 in safe arch, which means probably x86_64 doesn't have them at all? We'd have to always use a `for` loop I guess.

PR #121, opened 2022-12-30
Fix 120

Lokathor:
Closes https://github.com/Lokathor/wide/issues/120


Lokathor:
@CryZe I'm not clear why the wasm runs think that a file is missing, does it look like anything you've seen before?

CryZe:
Huh, no I've never seen that. You could try throwing an strace on the command to see what goes wrong maybe?

Lokathor:
I have no idea how to strace with wasmtime. I didn't add the wasm support in the first place >_>

maybe i can find someone on one of the discords.

PR #123, opened 2023-02-12
add saturating add/sub i16 and multiply and scale

mcroomp:
- Added saturating add and subtract for i16x8
- Added multiply and scale for i16x8 (equivalent of multiplying by a 16 bit fixed point between -1 and 1)
- Fixed two instances that mixed up sse and sse2 targets that caused syntax errors with target_features=-sse2

### DIFF
--- a/src/f64x2_.rs
+++ b/src/f64x2_.rs
@@ -438,7 +438,7 @@ impl f64x2 {
   #[must_use]
   pub fn fast_max(self, rhs: Self) -> Self {
     pick! {
-      if #[cfg(target_feature="sse")] {
+      if #[cfg(target_feature="sse2")] {
         Self { sse: max_m128d(self.sse, rhs.sse) }
       } else if #[cfg(target_feature="simd128")] {
         Self {
@@ -495,7 +495,7 @@ impl f64x2 {
   #[must_use]
   pub fn fast_min(self, rhs: Self) -> Self {
     pick! {
-      if #[cfg(target_feature="sse")] {
+      if #[cfg(target_feature="sse2")] {
         Self { sse: min_m128d(self.sse, rhs.sse) }
       } else if #[cfg(target_feature="simd128")] {
         Self {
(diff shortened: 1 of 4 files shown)


src/i16x16_.rs
```diff
@@ -662,6 +662,130 @@ impl i16x16 {
...
+  /// - _mm256_mulhrs_epi16 on avx2
+  /// - emulated via mul_i16_* on sse2
+  #[inline]
+  #[must_use]
+  pub fn mul_scale_round(self, rhs: Self) -> Self {
+    pick! {
+      if #[cfg(target_feature="avx2")] {
+        Self { avx2: mul_i16_scale_round_m256i(self.avx2, rhs.avx2) }
+      } else if #[cfg(target_feature="sse2")] {
+        // unfortunately mul_i16_scale_round_m128i only got added in sse3
```
Lokathor, line 744: another branch can be added here for sse3 if you'd like.

mcroomp: ok... it's for ssse3. not sure why there is also sse3.

src/f64x2_.rs
```diff
@@ -438,7 +438,7 @@ impl f64x2 {
   #[must_use]
   pub fn fast_max(self, rhs: Self) -> Self {
     pick! {
-      if #[cfg(target_feature="sse")] {
+      if #[cfg(target_feature="sse2")] {
```
mcroomp, line 441: This wouldn't compile otherwise with just sse set the wrong fields are declared

Lokathor: Ah, yeah there's probably not much testing in the crate for `sse` but *not* `sse2`. I think that in the long term `f64x2` should have its definition adjusted so that it uses an `m128d` field when just `sse` is set, but that can wait for later since it's kinda unrelated to this PR.

Lokathor:
I can merge now if you think it's all set.

mcroomp:
Awesome! good to go on my part

PR #124, opened 2023-02-13
Add aarch64 support along with emulator build as action

mcroomp:
Added neon intrinsics support for aarch64, along with emulator for testing.

Don't support armv7 at this point since it is unstable.


Lokathor:
This looks good so far, my main comment would be that newly added function/methods should generally be marked as `#[inline]` and also `#[must_use]`

mcroomp:
> 

You mean the Default and PartialEq trait? Other than that I don't think I added any functions

Lokathor:
Yep, even trait methods need to be specifically marked with `#[inline]` to make them eligible for inlining across codegen units when there's no generics.

Lokathor:
Yeah that makes sense.

Lokathor (approved):
Looks good here. Should I merge or is there even more to do?

mcroomp:
I think this is good to go now. I merged some conflicts with the previous change and also made it only use neon on aarch64, since armv7 also supports neon but causes compile issues since it is unstable.

mcroomp:
Thank you! Let me know when you publish the crate... can't wait to start using the new features :)

Lokathor:
I think we'll wait a little longer to sort out https://github.com/Lokathor/wide/pull/127 and then things should be fine to publish.

PR #126, opened 2023-02-16
Fix endianness in sqrt and add emulated MIPS target to test

mcroomp:


mcroomp:
Took a bit of fiddling to get the MIPS target building right with the emulator. Other people are using docker, but I thought it would be simpler to use the built-in options

mcroomp:
Not sure why the wasm test failed... seems to happen randomly

CryZe:
https://github.com/Lokathor/wide/actions/runs/4192493124/jobs/7268142870#step:4:9

> Run curl https://wasmtime.dev/install.sh -sSf | bash
  Installing latest version of Wasmtime ()
    Checking for existing Wasmtime installation
    Fetching archive for Linux, version 
https://github.com/bytecodealliance/wasmtime/releases/download//wasmtime--x86_64-linux.tar.xz 
curl: (22) The requested URL returned error: 404

It seems like it builds an invalid URL. Probably worth reporting upstream.

mcroomp:
> https://github.com/Lokathor/wide/actions/runs/4192493124/jobs/7268142870#step:4:9
> 
> > Run curl https://wasmtime.dev/install.sh -sSf | bash
> > Installing latest version of Wasmtime ()
> > Checking for existing Wasmtime installation
> > Fetching archive for Linux, version
> > https://github.com/bytecodealliance/wasmtime/releases/download//wasmtime--x86_64-linux.tar.xz
> > curl: (22) The requested URL returned error: 404
> 
> It seems like it builds an invalid URL (it's missing the version number). Possibly worth reporting upstream.

The shell script pulls https://api.github.com/repos/bytecodealliance/wasmtime/releases/latest and then tries to get the last tag to download the release. Looks like there's some kind of transient failure and it got a blank tag back, which then it tried to append to the URL which failed.

The only stable solution I see would be to hardcode instead of polling for the latest if this is a github API issue.

mcroomp:
Ok, so I added the wasmtime version to the test matrix, which I think is a reasonable solution. That will keep it from having a random dependency on querying github apis for the version which seems to fail sometimes (possibly a rate limit)

Lokathor (approved):

PR #127, opened 2023-02-20
made MulScaleRound a trait

mcroomp:
If you haven't published the new crate yet, I realized it woudl be better to make MulScaleRound a trait instead of a function since neon has a version that takes only a scalar. This would be a breaking change but if you haven't published a new version, it would be great if you could take this. Thanks!


Lokathor:
I haven't released the previous version yet, but can we rearrange this to have both the inherent method form, and then a trait which calls the inherent method? I'd like for people using a specific type like `i16x8` to not have to also pull in a trait, if possible.

the outline would be like this:
```rust
impl simd_type {
  pub fn do_math(self, rhs: Self) -> Self { todo!() }
}

impl MathTrait for simd_type {
  fn do_math(self, rhs: Self) -> Self {
    simd_type::do_math(self, rhs)
  }
}
```

If this feels like overkill then we don't have to, but it seems nicer to users.

mcroomp:
Yeah seems like overkill to have a trait just to overload a method. Just made into a normal method to be more like the other.

PR #135, opened 2023-09-12
add reduce_add to integer 

mcroomp:
Didn't add intrinsic versions since the compiler does a great job of vectorizing


Lokathor:
This seems fine, and don't worry about the mips CI failures, but you should put a doc comment on each of these new methods explaining what they do. I don't think the name is obvious enough on its own to skip a doc comment.

Lokathor (approved):

PR #136, opened 2023-09-17
Add dot product for i16. 

mcroomp:


mcroomp:
Temporarily disabled the nightly MIPS tests until this is fixed in the compiler

CryZe:
Do you have an upstream ticket for that? If it's indeed an upstream bug, then I'd argue it's very unlikely to be fixed, considering MIPS is basically unmaintained. And without an upstream ticket, even less so. It probably makes sense to rip out this from the CI entirely then. Was it there to test big endian? If so, it should probably be replaced with the much more maintained powerpc targets.

mcroomp:
Yeah the MIPS is just to test for endianness.... first thing that came to mind :)

I was going to try to make a simpler repro to file a ticket when I get some time.

Lokathor:
Yeah, MIPS got in the CI as "a big endian target", and doesn't otherwise have a person working on `wide` that knows any details about it.

mcroomp:
> Do you have an upstream ticket for that? If it's indeed an upstream bug, then I'd argue it's very unlikely to be fixed, considering MIPS is basically unmaintained. And without an upstream ticket, even less so. It probably makes sense to rip out this from the CI entirely then. Was it there to test big endian? If so, it should probably be replaced with the much more maintained powerpc targets.

Submitted issue... looks like 64 bit shifts in MIPS are broken.
https://github.com/rust-lang/rust/issues/116177

PR #147, opened 2024-01-16
feat:  add u8x16::unpack_low/high

RRRadicalEdward:
This PR adds `u8x16::unpack_low` and `u8x16::unpack_high` which is basically SSE2's `unpack_low_i8_m128i` and `unpack_high_i8_m128i`, but also allows _WASM_ _SIMD128_ and _ARM_ _NEON_ backends.

### DIFF
--- a/src/u8x16_.rs
+++ b/src/u8x16_.rs
@@ -459,6 +459,62 @@ impl u8x16 {
     }
   }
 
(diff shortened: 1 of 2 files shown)


Lokathor:
released in `wide-0.7.14`

PR #148, opened 2024-01-26
feat: add u8x16::narrow_i16x8

RRRadicalEdward:
Also, ran cargo fmt and fixed clippy warnings (added missing _inline_ attributes)


Lokathor:
released in `wide-0.7.15`

PR #151, opened 2024-04-29
simplify i16x16::move_mask

mcroomp:
SSE2 can actually do the whole thing. About 30% faster overall.


Lokathor:
> Published wide v0.7.17 at registry `crates-io`

PR #152, opened 2024-05-07
Add unsigned_abs to i8,i16,i32,i64 to provide well defined behavior for abs

mcroomp:
Add equivalent unsigned_abs that will not fail and returns the proper unsigned integer type. It's possible to do something similar with bitmuck and casts but the code looks horrible. Also added abs to i64 although intel doesn't support it, the other platforms do.


Lokathor (approved):

Lokathor:
Published wide v0.7.18

PR #154, opened 2024-05-18
improve perf of reduce_min,reduce_max,reduce_add and add unsigned upcast from u16x8

mcroomp:
From some reason the optimizer isn't vectorizing the reduction operations and ends up emitting a bunch of slow scalar code.


Lokathor (approved):

Lokathor:
published in `0.7.20`

PR #158, opened 2024-06-01
fix u32 sign bugs

mcroomp:
Couple bugs in u32x8 and u32x4 where i32 operations were used instead. Also fixed tests that were testing i32x8 instead of u32x8.


Lokathor:
This looks good but I'll leave it open for a bit in case you spot anything else. I won't be able to merge and publish this until at least later today, possibly a few days from now

mcroomp:
Ok I'm done. I also added a simple pseudo random test based on the existing From and Into traits that is much cleaner and less disruptive than the previous change.

PR #161, opened 2024-06-05
Adds boolean methods to u32x4 and u32x8

### LINKED ISSUE #160: Add methods from i32x4 for u32x4
I see there are some methods like all or any that are implemented for i32x4 but not u32x4.

I could go ahead and implement them unless I am missing something.

andyquinterom:
Closes #160

### DIFF
--- a/src/u32x4_.rs
+++ b/src/u32x4_.rs
@@ -509,6 +509,42 @@ impl u32x4 {
   }
 
   #[inline]
+  #[must_use]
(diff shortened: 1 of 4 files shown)


Lokathor:
This looks good so far, but you should puy a doc comment on each of these which explains what's going on

Lokathor:
Yeah that's probably good.

Lokathor:
released `0.7.27`

PR #164, opened 2024-06-29
Adds conversion functions from integers to floats

Mottl:
This PR adds four conversion functions from integers to floats:
`f32x4::from_i32x4`
`f32x8::from_i32x8`
`f64x2::from_i32x4`
`f64x4::from_i32x4`

### DIFF
--- a/src/f32x4_.rs
+++ b/src/f32x4_.rs
@@ -4,13 +4,13 @@ pick! {
   if #[cfg(target_feature="sse")] {
     #[derive(Default, Clone, Copy, PartialEq)]
     #[repr(C, align(16))]
-    pub struct f32x4 { sse: m128 }
+    pub struct f32x4 { pub(crate) sse: m128 }
   } else if #[cfg(target_feature="simd128")] {
     use core::arch::wasm32::*;
 
     #[derive(Clone, Copy)]
     #[repr(transparent)]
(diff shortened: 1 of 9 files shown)


Lokathor:
Looks like the build errors are because you're trying to use nightly stuff.

Also, these should either be `From` impls, or they should have more descriptive names that help the reader understand why it's not just a normal From impl. eg: `f64x2::from_i32x4_lower2(x)` lets the reader understand that they're dropping half their lanes.

Mottl:
Thanks for comments! I chose to implement `From`.

Mottl:
Btw, is there a reason for aarch64 in `#[cfg(all(target_feature="neon",target_arch="aarch64"))]`?
(compare with `#[cfg(target_feature="sse2")]`).

mcroomp:
They aren't really simple From since they are potentially lossy... i32 -> f32

Mottl:
Yes, according to `From` description you are right. Changed from `From` impls to ordinary functions.

Mottl:
@Lokathor, does it look good now?

Lokathor (approved):

Lokathor:
Released `wide-0.7.26`

Sorry about the delay! last weekend got unexpectedly busy

PR #166, opened 2024-07-07
add some missing widening operations

mcroomp:
Add widening from u8 -> u16 and unsigned widening multiply from u16


Lokathor:
released `0.7.27`

PR #168, opened 2024-07-26
ci: Run on macos-latest

waywardmonkeys:
Since `macos-latest` is an Apple Silicon machine now, aarch64 CI can be run natively.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,6 +25,11 @@ jobs:
         - { target: x86_64-pc-windows-msvc, toolchain: stable, os: windows-latest }
         - { target: x86_64-pc-windows-msvc, toolchain: beta, os: windows-latest }
         - { target: x86_64-pc-windows-msvc, toolchain: nightly, os: windows-latest }
+        # aarch64
+        - { target: aarch64-apple-darwin, toolchain: "1.60", os: macos-latest }
+        - { target: aarch64-apple-darwin, toolchain: stable, os: macos-latest }
+        - { target: aarch64-apple-darwin, toolchain: beta, os: macos-latest }
+        - { target: aarch64-apple-darwin, toolchain: nightly, os: macos-latest }
         # wasm32
         - { target: wasm32-wasi, toolchain: 1.56.0, os: ubuntu-latest, wasmtime: v5.0.0 }
         - { target: wasm32-wasi, toolchain: stable, os: ubuntu-latest, wasmtime: v5.0.0 }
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
This fails locally for me, I suspect it will here as well.

.github/workflows/rust.yml
```diff
@@ -49,12 +53,12 @@ jobs:
       run: cargo build --target ${{ matrix.rust.target }}
 
     - name: Test with default CPU features + No Default Cargo Features
-      if: matrix.rust.target == 'i586-pc-windows-msvc' || matrix.rust.target == 'i686-pc-windows-msvc' || matrix.rust.target == 'x86_64-pc-windows-msvc' || matrix.rust.target == 'wasm32-wasi'
+      if: matrix.rust.target == 'i586-pc-windows-msvc' || matrix.rust.target == 'i686-pc-windows-msvc' || matrix.rust.target == 'x86_64-pc-windows-msvc' || matrix.rust.target == 'aarch64-apple-darwin' || matrix.rust.target == 'wasm32-wasi'
```
waywardmonkeys, line 56: These lines cover all of the systems in the matrix above, so I think they could probably go away entirely instead of just adding a new platform to the list here and in each below.

waywardmonkeys:
To save some clicking through, the failures are:

```
---- t_f32x4::impl_f32x4_sin_cos stdout ----
thread 't_f32x4::impl_f32x4_sin_cos' panicked at tests/all_tests/t_f32x4.rs:409:9:
Wanted sin(-6058) to be -0.8464933 but got -0.8464934
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- t_f32x8::impl_f32x8_sin_cos stdout ----
thread 't_f32x8::impl_f32x8_sin_cos' panicked at tests/all_tests/t_f32x8.rs:653:9:
Wanted sin(-6058) to be -0.8464933 but got -0.8464934


failures:
    t_f32x4::impl_f32x4_sin_cos
    t_f32x8::impl_f32x8_sin_cos
```

Lokathor:
I don't know how "close enough" those results are, but they seem maybe close enough? we could adjust the test threshold

waywardmonkeys:
Just changing the threshold on those 2 tests from  `0.00000006` by removing a zero is enough for them to pass.

.github/workflows/rust.yml
waywardmonkeys: Removed in #171.

waywardmonkeys:
I tightened up the tolerances a bit more than I indicated in a previous comment, added in a job for macOS with 1.59 (not 1.56), with a comment about why.

Should be ready for review now!

waywardmonkeys:
Updated to use 1.60 and remove a comment in anticipation of #172 landing.

Lokathor (approved):

PR #173, opened 2024-07-28
Adding support for serde

michaelciraci:
This adds an optional serde feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,11 @@ default = ["std"]
 # Activate `std` within the crate. Currently this gives a much faster `sqrt`
 # impl when an explicit hardware sqrt isn't available.
 std = []
+serde = ["dep:serde"]
 
 [dependencies]
 safe_arch = { version = "0.7", features = ["bytemuck"] }
+serde = { version = "1", default-features = false, optional = true }
 bytemuck = "1"
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


Lokathor (approved):

Cargo.toml
```diff
@@ -19,9 +19,11 @@ default = ["std"]
 # Activate `std` within the crate. Currently this gives a much faster `sqrt`
 # impl when an explicit hardware sqrt isn't available.
 std = []
+serde = ["dep:serde"]
```
Lokathor, line 22: oh we need to have this feature removed and then have the `serde` feature just be the optional dep only, at least based on our current MSRV we do.

waywardmonkeys: Or rebase forward since this feature was added in Rust 1.60 and that's the new MSRV :)

Cargo.toml
```diff
@@ -19,9 +19,11 @@ default = ["std"]
...
 
 [dependencies]
 safe_arch = { version = "0.7", features = ["bytemuck"] }
+serde = { version = "1", optional = true }
```
waywardmonkeys, line 26: Should be safe to do a `default-features = false` in here as well.

Cargo.toml
michaelciraci: I just rebased so hopefully we should now be good

michaelciraci:
Rebased with new CI changes

Lokathor:
released `0.7.27`

PR #175, opened 2024-07-31
docs: Add missing backticks, enable `clippy::doc_markdown`

waywardmonkeys:

PR #176, opened 2024-08-02
fix CI problems by bumping msrv

Lokathor:
This isn't strictly necessary since the problem is that we're running against the MSRV of a dev-dependency and we could probably just replace that dev dependency, but I don't see a strong need to stick to the oldest possible minimum rust in this case.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,27 +11,27 @@ jobs:
       matrix:
         rust:
         # x86 without sse/sse2 on by default
-        - { target: i586-pc-windows-msvc, toolchain: "1.60", os: windows-latest }
+        - { target: i586-pc-windows-msvc, toolchain: "1.61", os: windows-latest }
         - { target: i586-pc-windows-msvc, toolchain: stable, os: windows-latest }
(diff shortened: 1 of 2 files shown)


Lokathor:
I'm gonna ping @CryZe and @waywardmonkeys on this one, what's up with these aarch64 link errors? Those seem like basic C functions, do we have something configured way wrong for those CI variants?

waywardmonkeys:
We could also just check in a `Cargo.lock` and keep it at the "working" versions.

As for the Aarch64 Linux thing, I was hoping they were just transient errors as they didn't happen before. I'll try and dig into it this weekend if I can.

waywardmonkeys:
Wonder if it only happens on nightly. Maybe we could add a `fail-fast: false` to the strategy to force it to not cancel when one fails ...

There's been a lot of bugs over the years apparently with Rust and these functions.

CryZe:
I believe I read that a recent compiler-builtins update accidentally changed the weakness of those functions, I think they are reverting it or so.

waywardmonkeys:
I re-pushed my PR #175 and that shows the issues with nightly and compiler builtins are fixed. Just MSRV now!

Issue #177, opened 2024-08-12
Compile error on latest version (0.7.27)


pthurston:
Hello, 

It looks like there is a compile error on the latest version (0.7.27). 

Example Cargo.toml

> [package]
> name = "wide-compile-issue"
> version = "0.1.0"
> edition = "2021"
> 
> [dependencies]
> wide = {version="0.7.27"}

Example src/main.ts
> fn main() {
>     println!("Hello, world!");
> }

Output from `cargo run`

> Blocking waiting for file lock on build directory
>    Compiling bytemuck v1.16.3
>    Compiling safe_arch v0.7.2
>    Compiling wide v0.7.27
> error[E0425]: cannot find function `convert_to_i16_m128i_from_lower8_u8_m128i` in this scope
>    --> /Users/pthurston/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wide-0.7.27/src/u16x16_.rs:243:27
>     |
> 243 |           a: u16x8 { sse: convert_to_i16_m128i_from_lower8_u8_m128i(v.sse) },
>     |                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a function with a similar name exists: `convert_to_i16_m128i_from_lower8_i8_m128i`
>     |
>    ::: /Users/pthurston/.cargo/registry/src/index.crates.io-6f17d22bba15001f/safe_arch-0.7.2/src/x86_x64/sse4_1.rs:177:1
>     |
> 177 | pub fn convert_to_i16_m128i_from_lower8_i8_m128i(a: m128i) -> m128i {
>     | ------------------------------------------------------------------- similarly named function `convert_to_i16_m128i_from_lower8_i8_m128i` defined here
> 
> error[E0425]: cannot find function `convert_to_i16_m128i_from_lower8_u8_m128i` in this scope
>    --> /Users/pthurston/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wide-0.7.27/src/u16x16_.rs:244:27
>     |
> 244 | ...u16x8 { sse: convert_to_i16_m128i_from_lower8_u8_m128i(unpack_high_i64_m128i...
>     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: a function with a similar name exists: `convert_to_i16_m128i_from_lower8_i8_m128i`
>     |
[...]

PR #179, opened 2024-08-17
Fix the i686 bugs

Lokathor:


Lokathor:
Oh good, it fails because `rand` won't build.

We're getting rid of that stupid dev-dep

mcroomp:
> Oh good, it fails because `rand` won't build.
> 
> We're getting rid of that stupid dev-dep

Added a PR for that #180

Lokathor:
Closing this on account of the i686 bugs were because Debian was doing extra silly things. https://github.com/Lokathor/wide/issues/177

PR #180, opened 2024-08-20
Remove Rand crate

mcroomp:
Causing dependency problems. Replace with simple LCG since it is just used for generating test value.

Also moved non_upper_case_globals into macro so that the visual studio code analyzer doesn't complain and generate bogus warnings.


Lokathor (approved):
I think with this we could set our MSRV back lower, but that's for another day, if anyone really cares (I don't think anyone cares about MSRV that far back).

PR #182, opened 2024-10-20
Added widening multiply for u16x8, u32x4, u32x8, i32x4 and i32x8

mcroomp:
Add implementations for all the intrinsics @Melirius might be interested

Also ran cargo fmt +nightly, so some comments got wrapped

### DIFF
--- a/src/i32x4_.rs
+++ b/src/i32x4_.rs
@@ -491,11 +491,16 @@ impl i32x4 {
     }
   }
 
+  /// Multiplies corresponding 32 bit lanes and returns the 64 bit result
+  /// on the corresponding lanes.
+  ///
+  /// Effectively does two multiplies on 128 bit platforms, but is easier
+  /// to use than wrapping mul_widen_i32_odd_m128i individually.
   #[inline]
   #[must_use]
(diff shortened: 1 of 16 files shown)


src/i32x4_.rs
```diff
@@ -490,6 +490,45 @@ impl i32x4 {
       }
     }
   }
+
+  #[inline]
+  #[must_use]
+  pub fn mul_widen(self, rhs: Self) -> i64x4 {
```
Lokathor, line 496: could we put a doc comment here about what happens, like the others have.

Lokathor (changes requested):

src/i32x4_.rs
mcroomp: Hi there, let me know if there is anything else in the PR. Thanks!

Lokathor (approved):

Lokathor:
> Published wide v0.7.29 at registry `crates-io`

mcroomp:
Don't forget to merge :)

Lokathor:
i didn't merge it!! :0

Lokathor:
okay now `0.7.30` should have it

PR #183, opened 2024-11-29
implement floor and ceil for f64 and f32 types

kralverde:
This PR implements the `floor` and `ceil` functions for the f64 and f32 types.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -43,7 +43,7 @@ jobs:
         target:  ${{ matrix.rust.target }}
 
     - name: Install wasmtime
-      if: matrix.rust.target == 'wasm32-wasi'
+      if: matrix.rust.target == 'wasm32-wasi' || matrix.rust.target == 'wasm32-wasip1'
       run: |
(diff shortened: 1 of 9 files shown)


kralverde:
I added test cases, but I'm not sure how to actually test on relevant hardware... there doesn't seem to be any github hooks or anything

Lokathor (approved):

CryZe:
The WASI target is now called `wasm32-wasip1`

kralverde:
> The WASI target is now called `wasm32-wasip1`

Thanks this worked, but now it looks like there's some kind of permission issue with the workflows...

kralverde:
Now an architecture error... I'm not super familiar with wasm if anyone wants to pitch in

Lokathor:
I'm gonna merge this, the wasm stuff shouldn't be on your shoulders.

@CryZe can you fix the wasm ci setup at some point later?

CryZe:
The problem is that it's still the old target name here (and possibly other places): https://github.com/Lokathor/wide/blob/main/.cargo-ci/config#L11

PR #184, opened 2024-12-03
disable wasm in ci for now

Lokathor:
just disable wasm for now


Lokathor:
i made this before i saw you posted that, see also https://github.com/Lokathor/wide/pull/185

PR #185, opened 2024-12-03
make wasm run in ci again

Lokathor:


.cargo-ci/config
```diff
@@ -8,6 +8,6 @@
 rustflags = ["-Ctarget-cpu=native"]
 rustdocflags = ["-Ctarget-cpu=native"]
 
-[target.wasm32-wasi]
+[target.wasm32-wasip1]
```
CryZe, line 11: You probably need to have the same target config for both in case you want to test the MSRV still.

.github/workflows/rust.yml
```diff
@@ -31,10 +31,10 @@ jobs:
         - { target: aarch64-apple-darwin, toolchain: beta, os: macos-latest }
         - { target: aarch64-apple-darwin, toolchain: nightly, os: macos-latest }
         # wasm32
-        #- { target: wasm32-wasi, toolchain: "1.61", os: ubuntu-latest, wasmtime: v5.0.0 }
-        #- { target: wasm32-wasi, toolchain: stable, os: ubuntu-latest, wasmtime: v5.0.0 }
-        #- { target: wasm32-wasip1, toolchain: beta, os: ubuntu-latest, wasmtime: v5.0.0 }
-        #- { target: wasm32-wasip1, toolchain: nightly, os: ubuntu-latest, wasmtime: v5.0.0 }
+        - { target: wasm32-wasip1, toolchain: "1.61", os: ubuntu-latest, wasmtime: v5.0.0 }
+        - { target: wasm32-wasip1, toolchain: stable, os: ubuntu-latest, wasmtime: v5.0.0 }
```
CryZe, line 35: Here without p1 now :D

CryZe:
Stable should be on p1 too.

Lokathor:
Is that something we can fix here or should we adjust the CI flow to bypass it and file an issue with rust-lang/rust?

CryZe:
Why would we file an issue upstream? You just need to pass the new P1 env vars as well.

Lokathor:
I misunderstood. I thought you were saying the target profile needed fixing, not just out environment variables.

PR #187, opened 2025-01-04
fix build on pre-sse2 CPUs

ceamac:
Build fails on Pentium 3


Lokathor:
released in `wide-v0.7.31`

PR #188, opened 2025-01-07
fix serialize impl of SIMD types

robinhundt:
change serialization of SIMD types to use serde::Serializer::serialize_tuple

The Serialize implementation of all types used Serializer::serialize_seq whereas the deserialize implementation delegated to that of the corresponding [T; LEN] type, which is implemented by serde using serialize_tuple. Formats like bincode will emit the lenght of the sequence but not of a tuple, leading to wrong deserialization.

Previously the following code would fail:
```rust
let ser = dbg!(bincode::DefaultOptions::new()
    .serialize(&u8x16::ZERO)
    .unwrap());
let der: u8x16 = bincode::DefaultOptions::new().deserialize(&ser).unwrap();
```
Output:
```
bincode::DefaultOptions::new().serialize(&u8x16::ZERO).unwrap() = [
    16,
    0,
    // 15 more zeros
]
thread 'block::tests::failing_deser' panicked at seec-core/src/block.rs:104:75:
called `Result::unwrap()` on an `Err` value: Custom("Slice had bytes remaining after deserialization")
```

and the following shows the difference between serialization and deserialization:
```rust
let ser = dbg!(bincode::serialize(&u8x16::ZERO).unwrap());
let der: u8x16 = dbg!(bincode::deserialize(&ser).unwrap());
```
Output:
```
[
    16,
    0,
    // 15 more zeros
]
[seec-core/src/block.rs:107:26] bincode::deserialize(&ser).unwrap() = (16, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0)
```

where we can see that the serialized data contains the length of the sequence which ends up in the deserialized vector type. In this case, deserialization does not panic because of the difference between `bincode::DefaultOptions` and the free functions in regards to trailing data.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,3 +26,6 @@ serde = ["dep:serde"]
 safe_arch = { version = "0.7", features = ["bytemuck"] }
 serde = { version = "1", default-features = false, optional = true }
 bytemuck = "1"
+
+[dev-dependencies]
+bincode = { version = "1.3.3" }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,7 +47,7 @@ use safe_arch::*;
 use bytemuck::*;
 
 #[cfg(feature = "serde")]
-use serde::{ser::SerializeSeq, Deserialize, Serialize};
+use serde::{ser::SerializeTuple, Deserialize, Serialize};
 
 #[macro_use]
 mod macros;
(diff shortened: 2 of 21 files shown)


Lokathor:
Would you consider this a "patch" update since the serialize/deserialize round trip never worked to begin with?

Lokathor:
Also, can you add some test cases to ensure that the round trip does work and that we don't break it accidentally in the future?

robinhundt:
I would probably consider it a patch update, but I'm not a 100% confident about any ramifications on other Ser/De implementations.

As I understand the serde data model and ecosystem, the current implementation is bugged if serde sequences and tuples are serialized differently, but worked if they were serialized the same. I **think** my fix should not impact serializers that worked before (e.g. serde-json where the sequences and tuples are the same).

I can add a test case but am unsure which serializer implementation to use. I personally like bincode and encountered the error with this format. Bincode only has 1 dependency which is serde. Should I add a test using bincode or would you prefer another format?

Lokathor:
Yeah use `bincode`, the one that had the problem. If you add it as a dev-dependency it will only be used for tests and stuff so it won't affect users of this crate as just a lib.

robinhundt:
I've added a test case round-tripping each SIMD type with bincode.

Lokathor (approved):

Lokathor:
released `wide-0.7.32`

PR #189, opened 2025-01-28
Use wrapping shifts in fallback implementations

imartayan:
The current fallback implementations of left/right shifts for integers use `<<`/`>>`, which panics when the shift amount is greater than the lane width (e.g. shifting a `u32x4` by 32).
Using `wrapping_shl`/`wrapping_shr` instead would solve this issue and match the behavior of the vectorized versions, which only keep the low bits of the shift amount (see e.g. https://doc.rust-lang.org/core/arch/wasm/fn.u32x4_shr.html).

### DIFF
--- a/src/i16x8_.rs
+++ b/src/i16x8_.rs
@@ -301,14 +301,14 @@ macro_rules! impl_shl_t_for_i16x8 {
           } else {
             let u = rhs as u64;
             Self { arr: [
-              self.arr[0] << u,
-              self.arr[1] << u,
-              self.arr[2] << u,
-              self.arr[3] << u,
-              self.arr[4] << u,
-              self.arr[5] << u,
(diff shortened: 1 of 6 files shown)


mcroomp:
These are a bit tricky since the SIMD shifts don't behave like wrapping shifts. For example, SIMD shift of a 1u32 << 32 will be zero, but for intel will be 1, since wrapping masks, but the SIMD wraps dont. So maybe panics are the right thing to do, maybe its better toto add a wrapping_shl that masks on SIMD and does not panic.

Lokathor:
i think if the "normal" effect people get, by which i mean the sse2 code path, never panics, then we should try to make all code paths never panic.

they don't need to be perfectly cross platform when weird inputs are given, but at least not a panic.

PR #190, opened 2025-02-15
Add `u8x32`

wingertge:
The lack of `u8x32` seems to be to just be an oversight, so I added it. If there's a reason it's not supported, please let me know.


Lokathor:
definitely just a "no one asked for it", happy for the inclusion

PR #202, opened 2025-07-11
Add variable bit shift for i64x2, i64x4, u64x2, u64x4

leung-ming:
Mostly follow #159 
"shr_each_i64_*" are only available on avx512 and I found https://stackoverflow.com/questions/31879878/how-can-i-perform-arithmetic-right-shift-in-c-in-a-portable-way
Should we implement it like above or let the compiler to decide what to do?

### DIFF
--- a/src/i64x2_.rs
+++ b/src/i64x2_.rs
@@ -248,7 +248,7 @@ impl Shl for i64x2 {
         // mask the shift count to 63 to have same behavior on all platforms
         let shift_by = rhs & Self::splat(63);
         Self { sse: shl_each_u64_m128i(self.sse, shift_by.sse) }
-      } else if #[cfg(all(target_feature="neon",target_arch="aarch64"))] {
+      } else if #[cfg(all(target_feature="neon", target_arch="aarch64"))] {
         unsafe {
           // mask the shift count to 63 to have same behavior on all platforms
           let shift_by = vandq_s64(rhs.neon, vmovq_n_s64(63));
@@ -306,7 +306,7 @@ impl Shr for i64x2 {
   #[inline]
   fn shr(self, rhs: Self) -> Self::Output {
     pick! {
-      if #[cfg(all(target_feature="neon",target_arch="aarch64"))] {
+      if #[cfg(all(target_feature="neon", target_arch="aarch64"))] {
         unsafe {
           // mask the shift count to 63 to have same behavior on all platforms
           // no right shift, have to pass negative value to left shift on neon
--- a/src/u64x2_.rs
+++ b/src/u64x2_.rs
@@ -248,7 +248,7 @@ impl Shl for u64x2 {
         // mask the shift count to 63 to have same behavior on all platforms
         let shift_by = rhs & Self::splat(63);
         Self { sse: shl_each_u64_m128i(self.sse, shift_by.sse) }
-      } else if #[cfg(all(target_feature="neon",target_arch="aarch64"))] {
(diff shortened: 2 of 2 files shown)


leung-ming:
Interesting CI fail, it seems that avx2 feature doesn't imply avx feature on windows? Or there are some CPUs or settings that make avx2 available but not avx?
Should I check both avx2 and avx here?

leung-ming:
Oh that's because it target to x86 arch.

leung-ming:
> "shr_each_i64_*" are only available on avx512 and I found https://stackoverflow.com/questions/31879878/how-can-i-perform-arithmetic-right-shift-in-c-in-a-portable-way Should we implement it like above or let the compiler to decide what to do?

It looks like the compiler does know what to do:
```bash
# After marking the function as never inline
$ RUSTFLAGS='-C target-feature=+avx2' cargo asm --release --lib "<wide::i64x4_::i64x4 as core::ops::bit::Shr>::shr"
    Finished `release` profile [optimized] target(s) in 0.03s
```
```asm
.section ".text.<wide::i64x4_::i64x4 as core::ops::bit::Shr>::shr","ax",@progbits
        .globl  <wide::i64x4_::i64x4 as core::ops::bit::Shr>::shr
        .p2align        4
.type   <wide::i64x4_::i64x4 as core::ops::bit::Shr>::shr,@function
<wide::i64x4_::i64x4 as core::ops::bit::Shr>::shr:
        .cfi_startproc
        mov rax, rdi
        vmovdqa ymm0, ymmword ptr [rsi]
        vpbroadcastq ymm1, qword ptr [rip + .LCPI2_0]
        vpand ymm1, ymm1, ymmword ptr [rdx]
        vpbroadcastq ymm2, qword ptr [rip + .LCPI2_1]
        vpsrlvq ymm2, ymm2, ymm1
        vpsrlvq ymm0, ymm0, ymm1
        vpxor ymm0, ymm0, ymm2
        vpsubq ymm0, ymm0, ymm2
        vmovdqa ymmword ptr [rdi], ymm0
        vzeroupper
        ret
```

Tritonn204:
I have a different set of PRs open on `safe_arch` and `wide` that bring stable AVX512 support to this system with the same API, thanks to Rust 1.89.

Would this change the way you handle this PR, or do you think leaving it to the compiler would still be best? More of a curiosity-based question than a suggestion.

leung-ming:
The compiler also say we should use AVX512, even though we don't have the 256-bit one.

target-feature=+avx512f
```asm
        mov rax, rdi
        vpbroadcastq ymm0, qword ptr [rip + .LCPI0_0]
        vpand ymm0, ymm0, ymmword ptr [rdx]
        vmovdqa ymm1, ymmword ptr [rsi]
        vpsravq zmm0, zmm1, zmm0 ; <--------------------   512-bit instruction
        vmovdqa ymmword ptr [rdi], ymm0
        vzeroupper
        ret
```

target-feature=+avx512vl
```asm
        mov rax, rdi
        vmovdqa ymm0, ymmword ptr [rdx]
        vpandq ymm0, ymm0, qword ptr [rip + .LCPI0_0]{1to4}
        vmovdqa ymm1, ymmword ptr [rsi]
        vpsravq ymm0, ymm1, ymm0
        vmovdqa ymmword ptr [rdi], ymm0
        vzeroupper
        ret
```

Lokathor:
Ah, very well then. We can always update more later, but i usually think it's better to merge some progress instead of no progress.

PR #213, opened 2025-09-10
fix: replace deprecated rustfmt setting and fix formatting

smu160:
## Summary 
This PR fixes the deprecated options warning when formatting code via `cargo +nightly fmt --all`, as seen below.

```
cargo +nightly fmt --all
Warning: the `merge_imports` option is deprecated. Use `imports_granularity="Crate"` instead
Warning: the `merge_imports` option is deprecated. Use `imports_granularity="Crate"` instead
Warning: the `merge_imports` option is deprecated. Use `imports_granularity="Crate"` instead
```

In addition, this PR includes all source code that was reformatted via `cargo +nightly fmt --all`.

The `rustfmt.toml` config file was updated in accordance with the aforementioned warnings.


Lokathor:
Ah, okay neat. If you just utilize "format on save" and not manual calls to rustfmt you never notice the warnings about old settings.

Issue #206, opened 2025-08-26
Masks should be unsigned


RagnarGrootKoerkamp:
I've been using `wide` recently ([this commit](https://github.com/RagnarGrootKoerkamp/sassy/commit/8c5fbd4136bbde421bccfc4a6cfb6a24131273b4)), and ran into a number of small things. Didn't bother making issues since I managed to work around everything, but since I see that you just pushed 1.0 (#204), let me make some anyway.

First up, `move_mask` ought to return a `u32` instead of `i32`. Currently, something like `eq.move_mask() as u64` is broken because of the (IMO) unexpected sign extension that treats the top-bit of the mask in a special way.

RagnarGrootKoerkamp:
If we're changing types anyway, a second consideration is whether they should have the same width as the number of lanes. With `u32` you can never support `u8x64` for AVX512.

On the other hand, for 2 or 4 lanes, you would still return an 8-bit `u8` so it's never quite perfect.

PR #216, opened 2025-09-13
movemask update

Lokathor:
Closes https://github.com/Lokathor/wide/issues/206


Lokathor:
@RagnarGrootKoerkamp this is what you had in mind?

RagnarGrootKoerkamp:
I have some more time now -- I might try to make some bigger PRs as well to make things consistent between the integer types.

Lokathor:
won't have much time to code today it turns out. i'd rather merge this breaking change now and then more move masks can be added as non-breaking updates if people do need those after all.

PR #229, opened 2025-11-03
to_bitmask implemented for all unsigned ints.

tower120:
There are no tests for `i16x32` and `i32x16` - because I don't know what exactly they should be tested for.
I understand what they should do, but your tests for other SIMDs are more sophisticated then my understanding. I don't see what they're actually testing, what corner cases.


tower120:
I think that error 
https://github.com/Lokathor/wide/actions/runs/19052661770/job/54416176485?pr=229
from your side...

Lokathor:
Ah, looks like a `safe_arch` bug for sure.

PR #233, opened 2025-11-08
Add support for `wasm64` with `simd128`

KonaeAkira:
Even though the `memory64` feature has been [standardized](https://webassembly.org/features/), the `wasm64-unknown-unknown` target is currently only a [tier 3 target](https://doc.rust-lang.org/rustc/platform-support/wasm64-unknown-unknown.html), so it's perfectly understandable if you don't want to merge this PR.

I stumbled upon this while experimenting with building my project with `wasm64-unknown-unknown`.


Lokathor:
I don't know that the build_check_only pass uses Nightly at all, so you'll probably have to change it to be a Nightly CI pass.

Lokathor:
I think it's the same safe_arch bug that keeps happening. I really have to look in to that.

PR #242, opened 2025-12-11
 Proper float negation; fixes #241 

### LINKED ISSUE #241: Floating-point negation is incorrect in edge cases, prevents compiler optimizations
`wide` currently implements sign negation on floats as `0.0 - x`: docs.rs/wide/1.0.2/src/wide/lib.rs.html#371-390

This is both semantically incorrect and slow.

Here's a playground link demonstrating the semantic difference between `0.0 - x` and true negation:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=99548e3611548e6fbcc1a185dde58642

NEON has a dedicated instruction for this, but x86 does not; the correct way to implement this on x86 is is bitwise XOR with a mask that only has the sign bits set.

The reason I care is not because of edge cases but because it inhibits compiler optimizations. For example, `a.mul_add(-b, c)` does not optimize into the negated fused multiply-add instruction; instead it emits a `vsubps` followed by a regular FMA instruction.

Shnatsel:
AI use disclosure: the change to `impl Neg` is handwritten, while the tests were written by Claude 4.5 Opus

### DIFF
--- a/tests/all_tests/t_f32x16.rs
+++ b/tests/all_tests/t_f32x16.rs
@@ -126,6 +126,30 @@ fn impl_sub_for_f32x16() {
 }
 
 #[test]
+fn impl_neg_for_f32x16() {
+  let a = f32x16::from([
+    1.0, -2.0, 3.0, -4.0, 5.0, -6.0, 7.0, -8.0, 0.0, -0.0, f32::INFINITY,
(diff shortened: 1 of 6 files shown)


Lokathor:
released in `1.0.3`

Shnatsel:
That was quick! Thanks!

PR #255, opened 2026-04-09
Add missing `f32x16` functionality and fix incorrect behaviour

Noam2Stein:
Notes:

My CPU does not support `avx512f` so i did not run tests with the feature enabled. Tests should probably be ran before merging.

Three functions are missing optimizations that require missing `safe_arch` intrinsics (see commit log).

While adding the missing functionality i found two `f32x16` bugs that i fixed, though this does mean behaviour is changed and a minor SemVer update may be needed:

The mask used in `is_finite` was incorrect and was causing infinite values to be seen as finite.

`round_int`, unlike its counterparts from other wide float types, did not handle NaNs or other edge cases (that is the job of `fast_round_int`).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # `wide` Changelog
 
+## Unreleased
+
+* Added floating-point constants to `f32x16` and made `f32x16::new` a `const fn`.
+
 ## 1.2.0
 
 * added reduce operations and dot to `i16x32`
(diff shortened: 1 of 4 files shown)


Lokathor:
tests are always ran before merging! well, sometimes they're skipped on docs-only changes, but code changes always get CI.

math bug fixes aren't considered a semver break, so no semver concerns there.

this all looks great.

Noam2Stein:
> this all looks great.

Thank you! first contribution to a crate that is not mine.

PR #256, opened 2026-04-10
Re-export bytemuck and explain casting

nnmm:
This solves two problems:
* It makes it more obvious how to cast between SIMD types
* It allows you to use the exact version of `bytemuck` that `wide` uses, without risking depending on a different version of `bytemuck` by accident

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,6 +19,16 @@
 //! instructions to complete the task. In the worst case, the code just becomes
 //! totally scalar (though the math is still correct, at least).
 //!
+//! ## Casting
+//!
+//! The SIMD types implement the [`bytemuck::Pod`] trait, which means that it
+//! is possible to do bitwise casts between SIMD types of the same size with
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -19,6 +19,16 @@
...
 //!
+//! ## Casting
+//!
+//! The SIMD types implement the [`bytemuck::Pod`] trait, which means that it
+//! is possible to do bitwise casts between SIMD types of the same size with
+//! the [`bytemuck::cast()`] function and others. `bytemuck` is re-exported by
+//! this crate for convenience.
+//!
+//! This typically does not have much, if any, runtime overhead in optimized
+//! builds.
```
nnmm, line 30: I used a vague formulation here because I don't know exactly. Should I leave this out?

Lokathor:
yeah. there definitely *can* be a (small) runtime cost for some casts, so "not much if any" is probably the most appropriate way to describe it.

PR #258, opened 2026-04-17
Add more wide-float functionality

Noam2Stein:
Added functions are: `trunc`, `fract`, `signum`, `midpoint`, `clamp`, `fast_clamp`, `reduce_mul`, `exp2`.

For clamp, because `min` and `max` already sacrifice performance to be consistent with
`std`, `clamp` is also consistent with `std` and has an assertion for `min <= max`. Because of
this there must also be a `fast_clamp` function that is faster.


Noam2Stein:
On second thought `clamp` having an assertion is likely not useful for most use cases. If you think the assertion should be removed i will remove it and `fast_clamp` because without an assertion it is identical to `clamp`.

Noam2Stein:
@Lokathor i am tagging you just in case you did not receive a notification from GitHub.

Lokathor:
I'll try to merge this soon, but I have had limited time lately. Definitely not forgotten though.

Lokathor:
This looks good, but for the `midpoint` methods would it be better to multiply by 0.5 instead of divide by 2.0? I know it's basically the same output, but then you get just one multiply instruction instead of a whole division sequence right? I'm not sure if that's one of those minor optimizations that llvm does on its own or that we should write into the code.

Noam2Stein:
I have always thought multiplication and division have the same cost since they both have their own instructions, but this can obviously be wrong. I will change it.

Also comment on what you think the behavior of `clamp` should be.

Lokathor:
Looking at some stack overflow discussions, it seems that a floating Mul will never be slower than a floating Div, and depending on architecture it will sometimes be faster. And so let's keep the mul.

For clamp, I think a debug_assert instead of an assert hits the sweet spot between checking and speed. Then we don't need two versions. Does that sound good?

Noam2Stein:
I also added `Rem`, `div_euclid` and `rem_euclid`.

It seems it is not possible to implement `Rem` with SIMD instructions. I tried to translate the implementation from `libm` but it has loops. Perhaps it is possible to use SIMD instructions for the non-loop parts of the implementation.

At least `div_euclid` and `rem_euclid` use SIMD instructions after calling `Rem`.

Noam2Stein:
Is there something to fix before this is merged? asking just in case.

Lokathor:
I just lost this a little down the pile. I'll try to release it soon.

PR #259, opened 2026-04-25
Rename float function `sign_bit` to `is_sign_negative` and add `is_sign_positive`

Noam2Stein:
The name `sign_bit` makes the behavior of the function unclear. `is_sign_*` are also used in `std::simd`.

For `f32` types i deprecated the original function. `f64` types did not expose the original function in the first place.

I put this in a separate PR because this involves deprecating functions and so you may not want it.

Note: if this is not merged for `v1.4.0` the `deprecated` attributes would need to be updated.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,10 @@
 # `wide` Changelog
 
+## Unreleased
+
+* Renamed float function `sign_bit` to `is_sign_negative` and added
+  `is_sign_positive`.
+
 ## 1.3.0
 
 * Fixes the behaviour of `f32x16` functions `is_finite` and `round_int`. They
--- a/src/f32x16_.rs
+++ b/src/f32x16_.rs
@@ -1273,9 +1273,12 @@ impl f32x16 {
   #[inline]
   #[must_use]
   pub fn is_sign_positive(self) -> Self {
-    let t1 = cast::<_, i32x16>(self);
-    let t2 = t1 >> 31;
-    cast::<_, f32x16>(t2).simd_eq(f32x16::ZERO)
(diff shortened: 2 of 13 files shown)


Lokathor:
A few things come to mind, which aren't entirely connected so I'll make a bullet list

* `sign_bit` definitely does *not* do what I would expect it to. I would have expected it to extract the bit pattern of the sign bit, unmodified. Instead, it's splatting the sign bit across all other bits.
* This should be documented on the function. Even if we depreciate the function, it should get a doc comment.
* I agree we should depreciate the function and make new ones with clearer names.
* ~~The actual math used to get the result seems to be sub-optimal. I think what we *want* is `self & -0.0` to get the sign bit value by itself, then we do either `tmp == -0.0` or `tmp != -0.0` to splat that value across the lane. Casting to int and bit shifting around is not as good. The simd units for int and float are not the same on many architectures, and so doing an int operation between several float steps will cause a slow down. Let's not count on LLVM to optimize it when we can make the code "proper" in the first place.~~

Lokathor:
Wait wait, on the last point, I guess equality testing with -0.0 doesn't work, because 0.0 is also equal to -0.0 under float equality, so we do need a bit equal test.

Noam2Stein:
About the implementation, it should be possible to do `self & -0.0` then integer equality. Bitand should be faster than left shift.

`f64` types previously did not expose `sign_bit`. Should i add it for them to match `f32` types? even if deprecated.

Lokathor:
Definitely we don't want to add a method we immediately deprecate. Having just the `is_sign_...` methods is fine.

PR #260, opened 2026-04-25
Add array operations

Noam2Stein:
Currently this adds an implementation for `Index`.

I originally also wanted to add `.map(|..| ..)` and `_::from_fn(|i| ..|)`. These functions are useful for using scalar functions that are not available for wide types, for example `some_f32x4.map(f32::sinh)`. Do you think that is a good idea?

The index implementation supports slices, like the index implementation for arrays.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # `wide` Changelog
 
+## Unreleased
+
+* Implemented `Index` for all types.
+
 ## 1.3.0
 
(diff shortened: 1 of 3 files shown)


Lokathor:
I don't think that having `Index` impls is a good idea. It leads to non-obvious performance foot guns. It should probably be very visually explicit in the code that you're moving from a wide type to an array, then doing some non-wide operation several times, then going back to a wide type.

PR #262, opened 2026-05-09
Add 6 new SIMD math functions + fix ln precision + extend exp range

CokieMiner:
## Overview

Six brand-new math functions across all 6 SIMD float types (`f32x4/8/16`, `f64x2/4/8`), plus precision and range fixes for the two existing functions (`exp`, `ln`) they depend on, plus IEEE 754-2019 compliant NaN propagation for `clamp`.

### New functions

| Function | All 6 widths | Algorithm |
|----------|-------------|-----------|
| `exp_m1` | x | Taylor poly (deg 13 f64 / deg 7 f32), `z*2^n + (2^n-1)` reconstruction |
| `ln_1p` | x | `x*ln(1+x)/((1+x)-1)` identity, mathematically exact |
| `sinh` | x | Taylor poly for `abs(x) < 0.5`, `(e^x-1/e^x)/2` otherwise |
| `cosh` | x | Taylor poly for `abs(x) < 0.5`, `(e^x+1/e^x)/2` otherwise |
| `tanh` | x | `-exp_m1(-2x)/(exp_m1(-2x)+2)`, small/large shortcuts |
| `cbrt` | x | SLEEF poly for `1/cbrt(d)`, 1 Newton on `1/cbrt`, 1 polish |

### Existing functions improved

| Function | Issue | Fix |
|----------|-------|-----|
| `exp` | Range clamped too early (87.3/708.4), subnormal inputs returned 0 | pow2n excess decomposition, subnormal `vm_pow2n`, underflow early-return |
| `exp2` | Range clamped too early, same pattern as `exp` | Same excess decomposition fix |
| `ln` | LN2 constants from f32-literals (10 sig digits) caused 630 ULP error in f64 | fdlibm bit-exact `f64::from_bits(0x3FE62E42FEE00000)` |
| `clamp` | No NaN propagation, unspecified behavior for `min > max` | IEEE 754-2019 compliant `clamp` with explicit NaN blend; old behavior preserved as `fast_clamp` |

### Infrastructure

| Component | Change |
|-----------|--------|
| `vm_pow2n` | Subnormal support via per-lane integer shift |
| `polynomial_7!` | New macro in `lib.rs` for f64 cosh Taylor evaluation |

## Precision: 1 ULP for all functions except cbrt (2 ULP)

Verified across all 6 SIMD widths via:
- 30 edge-case tests (overflow boundaries, subnormal bounds, +/-0.0, +/-inf, NaN)
- 2000-point LCG-random fuzz tests per function (per-lane ULP comparison vs scalar reference)
- 8-function NaN-propagation tests (lane 0 = NaN, lane 1+ = finite)
- Subnormal-specific `exp` tests for `x < -87` (f32) / `x < -708` (f64)
- Sign preservation tests for `exp_m1(-0.0)`, `tanh(-0.0)`, `cbrt(-0.0)`, `ln_1p(-0.0)`

| Function | f32 ULP | f64 ULP |
|----------|---------|---------|
| `exp` | 1 | 1 |
| `exp2` | 1 | 1 |
| `ln` | 1 | 1 (was ~630) |
| `exp_m1` | 1 | 1 |
| `ln_1p` | 1 | 1 |
| `sinh` | 1 | 1 |
| `cosh` | 1 | 1 |
| `tanh` | 1 | 1 |
| `cbrt` | 2 | 2 |

### SIMD branching strategy: `.all()` vs per-lane `blend`

Two patterns are used throughout:

**`if mask.all() { return ... }` : homogeneous fast path.** When all lanes hit the same trivial case (all negative, all zero, all large), skip the entire computation and return a constant. The branch predictor handles this: for mixed data it's predicted not-taken (~0 cycle overhead); for homogeneous edge batches it's taken (~0.5ns total). Used in `exp_m1` (neg-asymptote), `exp` (underflow), `tanh` (large-x), `cbrt` (zero).

**Per-lane `mask.blend(a, b)` : heterogeneous path.** When lanes may differ, compute both paths for all lanes and blend per-element. This is the standard SIMD pattern: SIMD can't branch per lane, so both code paths execute and the mask selects the correct result. Used for `sinh`/`cosh` (poly vs exp-based), `tanh` (small-x vs exp_m1), overflow/underflow blending.

### `exp` / `exp2` / `exp_m1` : pow2n excess decomposition

**Problem:** `vm_pow2n(r)` overflows when `r` exceeds the IEEE max normal exponent (127 for f32, 1023 for f64). The previous code clamped `max_x` at 87.3/708.4, losing the top ~1% of the representable range and silently returning inf for inputs like `exp(87.5)` or `exp(708.5)`. `exp2` had the same issue with a `|x| < 120` guard.

**Solution:** Keep the original unreduced argument `x` and original `r` from range reduction (preserving full precision). Decompose `2^r = 2^r_safe * 2^excess` where `r_safe <= max_r` and `excess in {0,1}`. Reconstruct as `(z+1) * 2^excess * 2^r_safe`. This avoids both intermediate overflow and the precision loss of LN2 pre-subtraction.

### `ln` : LN2F constant fix

**Problem:** `LN2F_HI` and `LN2F_LO` were sourced from f32-literals (e.g., `0.693359375`, `-2.12194440e-4`), providing only ~10 significant digits. In f64 `ln`, this produced ~630 ULP error. Cascaded through `ln_1p`, `log2`, `log10`, and `pow`.

**Solution:** Replaced with fdlibm's bit-exact double-double decomposition: `f64::from_bits(0x3FE62E42FEE00000)` and `f64::from_bits(0x3DEA39EF35793C76)`. f32 `ln` retains the f32-precision split which sums correctly to 32-bit `ln(2)`.

### `ln_1p` : identity-based precision

`ln(1+x)` for small `x` loses precision because `1+x` rounds. The standard fix uses `ln_1p(x) = x * ln(1+x) / ((1+x) - 1)`. When `u = 1+x` rounds, `u-1` loses bits, but the ratio `ln(u)/(u-1)` naturally compensates. When `u == 1` exactly, return `x` directly. This is the MUSL/fdlibm approach: mathematically exact, no threshold needed.

Edge cases tested: `ln_1p(-1) -> -inf`, `ln_1p(x < -1) -> NaN`, `ln_1p(-0.0) -> -0.0`.

### `tanh` : `exp_m1`-based with shortcuts

**Formula:** `tanh(x) = -exp_m1(-2x) / (exp_m1(-2x) + 2)`. Avoids both catastrophic cancellation (direct `(e^x - e^-^x)/(e^x + e^-^x)` near 0) and overflow (direct exp-based formula for large x).

**Small-x:** For `|x| < 5*10^-^8` (f64) / `2*10^-^4` (f32), `tanh(x) ~= x` because `x^3/3 < 16*ULP(x)`. Derivation: `x^2 < 48*2^-^p^-^1` -> `x < 1.03*10^-^7` / `2.39*10^-^3`. Thresholds have 2-10* safety margin.

**Large-x:** For `|x| > 9.011` (f32) / `|x| > 19.062` (f64), `tanh(x) = sign(x)` exactly (mantissa exhaustion: `2e^-^2^x < 2^-^p^-^1`). Derivation: `e^-^2^x < 2^-^p^-^1` at `x > 8.3` / `18.4`. Uses `if large.all() { return sign(x); }` early-return.

### `sinh` / `cosh` : Taylor crossover at 0.5

For `|x| < 0.5`, use Taylor polynomials (6 terms f64, 3 terms f32 for sinh; 7 terms f64 via `polynomial_7!`, 3 terms f32 for cosh). Beyond crossover, use `(e^x +/- 1/e^x)/2` via `a.exp()`. The crossover is chosen so the last truncation term is < 1 ULP at x=0.5 for both types.

### `cbrt` : SLEEF polynomial + two refinements

Polynomial approximation for `1/cbrt(d)` on [0.5, 1.0) using SLEEF's minimax coefficients. One Newton step on `1/cbrt`: `x = x - d*mul_add(x^4, -x) * (1/3)`. One final polish on `cbrt(d)`: `y = y - (2/3)*y*(y*x - 1)`. Exponent scaling decomposes into `k = floor(e/3)` and remainder adjustment via `cbrt(2)` / `cbrt(4)` constants. Subnormals are scaled up by `2^2^4` (f32) or `2^5^4` (f64), then scaled back by `2^8` / `2^1^8`.

Special values (NaN, +/-0.0, +/-inf) override the computed result via `blend` after `flip_signs` is applied -- the blend comes last so it takes priority and no value gets double-flipped.

2 ULP precision (not 1): accumulated roundoff in the `r = e - 3k` scaling step plus the two-refinement chain prevents a 1-ULP guarantee across all platforms.

### `vm_pow2n` : subnormal support

The standard bit-cast trick (`n + bias + 2^K` as float -> int -> shift -> float) constructs floats via their IEEE bit pattern. This only works for normal exponents. For subnormals, we detect `n < -126` (f32) / `n < -1022` (f64) and construct the mantissa directly: `mantissa = 1 << (n + offset)` where offset is 149 (f32) / 1074 (f64). Uses per-lane integer shift.

### `exp_m1(-0.0)` sign preservation

The reconstruction `z*2^n + (2^n-1)` produces `+0.0` for `x = -0.0`. Added `is_zero.blend(self, result)` to propagate the input sign.

### `clamp` / `fast_clamp` : IEEE 754-2019 compliance

`clamp` now computes `min(max, max(min, self))` then blends NaN over any lane where any input is NaN. `fast_clamp` preserves the original intrinsic-only behavior for hot paths where NaN is impossible.

### Early-return optimizations

Each uses `if mask.all() { return ... }` : predicted not-taken, ~0-cycle overhead for normal data, 4-26* speedup for homogeneous edge batches.

| Function | Trigger | Returns | Why |
|----------|---------|---------|-----|
| `exp_m1` | `x < -17.329` (f32) / `x < -37.429` (f64) | `-1.0` | `e^x < 2^-^p^-^1` : mantissa exhausted |
| `tanh` | `abs(x) > 9.011` (f32) / `abs(x) > 19.062` (f64) | `sign(x)` | `2e^-^2^x` below precision |
| `cbrt` | `x = +/-0.0` | `self` | preserves -0.0, skips poly+Newton |
| `exp` | `x < -103.63` (f32) / `x < -744.79` (f64) | `0.0` | `e^x` below minimum subnormal |

## Performance

All benchmarks on AMD Zen 5 (Ryzen AI 7 350) (AVX-512). Adaptive warmup (~200ms), 11-trial minimum-of-means. Mixed-lane inputs test branch divergence.

### Throughput : mixed lanes (ns/call)

| Function | f64x2 | f64x8 | f32x4 | f32x8 |
|----------|-------|-------|-------|-------|
| `exp` | 7.4 | 23.3 | 5.8 | 10.0 |
| `ln` | 4.4 | 20.8 | 5.3 | 9.4 |
| `exp_m1` | 9.1 | 27.8 | 7.9 | 13.8 |
| `ln_1p` | 5.9 | 26.0 | 5.3 | 12.0 |
| `sinh` | 10.1 | 31.7 | 7.4 | 12.3 |
| `cosh` | 10.0 | 31.3 | 7.1 | 11.7 |
| `tanh` | 14.5 | 41.9 | 11.2 | 19.7 |
| `cbrt` | 11.8 | 34.3 | 9.4 | 16.8 |

SIMD-vs-scalar ratios vary by function and CPU. On Zen 5, 8 independent `libm::exp` calls can pipeline in parallel via ILP, so SIMD `exp`/`ln` don't always win. Complex functions (`cbrt`) show consistent 1.3-2.1* SIMD advantage. On non-x86 (ARM/WASM/RISC-V) without hardware math acceleration, SIMD wins everywhere. Per-lane efficiency scales well: all functions under 7ns/lane at 8-wide.

### Early-return speedups (homogeneous edge inputs)

| Bench | SIMD | Scalar | Ratio |
|-------|------|--------|-------|
| `exp_m1` f64 `[-50,-40]` | 0.3ns | 2.2ns | 7.2* |
| `exp_m1` f32 `[-20..-100]` | 0.4ns | 9.3ns | 26.5* |
| `tanh` f64 `[50.0;-30.0]` | 0.5ns | 2.4ns | 4.6* |
| `tanh` f64x8 `[50.0;8]` | 2.0ns | 8.7ns | 4.3* |

## Threshold audit

Every numeric threshold is justified:

| Threshold | Value | Derivation | Source |
|-----------|-------|------------|--------|
| sinh/cosh crossover | 0.5 | Last Taylor term < 1 ULP at x=0.5 | Taylor analysis |
| tanh small (f64) | `5e-8` | `x^2 < 48*2^-^5^2` -> `x < 1.03e-7` | Derived |
| tanh small (f32) | `2e-4` | `x^2 < 48*2^-^2^3` -> `x < 2.39e-3` | Derived |
| tanh large (f64) | 19.062 | `e^-^2^x < 2^-^5^4` (mantissa exhaustion) | Derived |
| tanh large (f32) | 9.011 | `e^-^2^x < 2^-^2^4` (mantissa exhaustion) | Derived |
| exp_m1 neg-asymptote (f64) | -37.429 | `e^x < 2^-^5^4` -> `x < -37.429` | Derived |
| exp_m1 neg-asymptote (f32) | -17.329 | `e^x < 2^-^2^5` -> `x < -17.329` | Derived |
| exp max_x (f64) | 709.783 | `ln(f64::MAX) ~= 709.7827` | IEEE 754 |
| exp max_x (f32) | 88.723 | `ln(f32::MAX) ~= 88.7229` | IEEE 754 |
| exp min_x (f64) | -744.79 | `-1074.5*ln(2)` | IEEE 754 |
| exp min_x (f32) | -103.63 | `-149.5*ln(2)` | IEEE 754 |
| exp max_r (f64) | 1023 | IEEE max normal exponent | IEEE 754 |
| exp max_r (f32) | 127 | IEEE max normal exponent | IEEE 754 |
| ln LN2F_HI/LO (f64) | `from_bits(0x3FE6...)` | fdlibm double-double ln(2) | fdlibm |
| exp LN2D_HI/LO | `0.69314575... + 1.4286e-6` | fdlibm exp ln(2) split | fdlibm |
| cbrt C0..C5 | SLEEF coefficients | Minimax poly `1/cbrt(d)` on [0.5,1) | SLEEF |
| ln P/Q coeffs | fdlibm coefficients | Rational approx `ln(1+z)` | fdlibm |
| Taylor coeffs | `1/k!` | Exact mathematical constants | -- |

## Attribution

| Algorithm | Source | License |
|-----------|--------|---------|
| cbrt minimax polynomial | Naoki Shibata et al., SLEEF | Boost 1.0 |
| ln LN2F_HI/LO double-double | Sun Microsystems, fdlibm | Public domain |
| exp/exp_m1 range reduction | fdlibm | Public domain |
| ln_1p identity | Rich Felker et al., MUSL libc | MIT |

## Files changed

| File | Change |
|------|--------|
| `src/lib.rs` | Added `polynomial_7!` macro |
| `src/f32x4_.rs`, `f32x8_.rs`, `f32x16_.rs` | 6 new functions + `exp`/`exp2`/`ln` fixes + `vm_pow2n` subnormals + `clamp`/`fast_clamp` |
| `src/f64x2_.rs`, `f64x4_.rs`, `f64x8_.rs` | 6 new functions + `exp`/`exp2`/`ln` fixes + `vm_pow2n` subnormals + `clamp`/`fast_clamp` |
| `tests/all_tests/t_simd_float.rs` | 2000-point LCG fuzz, overflow/subnormal/NaN-propagation edge cases for all 8 functions |
| `benches/math_benchmarks.rs` | Adaptive throughput/homogeneous/boundary/latency/per-lane benchmarks |
| `Cargo.toml` | Added `[[bench]] math_benchmarks` |


CokieMiner:
ok my cbrt implementation needs fixes :|

CokieMiner:
Ok so, it was just a too strict test that failed cbrt precision checks, I think probably they don't have proper fma don't know, will keep documented as 2 ulp but on x86 hardware it should be 1ulp.

CokieMiner:
**Summary of improvements:**

*   **Fixed Fuzz Tests**: Resolved a bug in the `f32` test generator that was masking sign bits. Fuzz tests now correctly cover negative inputs for `sinh`, `tanh`, `cbrt`, etc.
*   **Extended `exp2` Range**: Refactored `exp2` to support the full representable range (up to $2^{127}$ for f32 and $2^{1023}$ for f64), matching the logic recently added to `exp`.
*   **Improved Coverage**: Updated `ln_1p` fuzzing to include the negative domain $x \in (-1, 0)$.
*   **Numerical Stability**: Adjusted test tolerances to handle subnormal precision edge cases without failing on bit-perfect results.

CokieMiner:
Just remembered I gave attribution in the PR description but didn't leave comments with proper attribution. I don't think there will be any licensing problem as the original code is in C not Rust and math can't be licensed. In case you want proper citation directly in comments just ask and I'll push a commit.

Lokathor:
Yeah let's have a citation within the code comments, that's a lot easier to see than finding a particular PR comment way back in the git history.

CokieMiner:
Sumary:
#### **1. Formal Algorithm Citations**
Added technical attributions and verified constants against upstream sources to ensure auditability:
*   **SLEEF**: Attributed `cbrt` minimax polynomials to Naoki Shibata et al. (Boost License 1.0).
*   **FDLIBM (Sun Microsystems)**: Verified bit-exact `ln(2)` double-double decompositions for $f64$ accuracy.
*   **MUSL / Netlib**: Documented high-precision identities used for $ln(1+x)$ rounding compensation.

#### **2. Tightened Asymptote Thresholds**
Changed thresholds to include more values in the snapping path to identity constants. ($\epsilon$).
*   **`tanh`**: Updated to **9.011** ($f32$) and **19.062** ($f64$).
*   **`exp_m1`**: Updated to **-17.329** ($f32$) and **-37.429** ($f64$).

#### **3. Enhanced Symmetry Testing**
*   Updated `cosh` unit tests to use **full-range signed random inputs** instead of absolute values.
*   Confirmed $\cosh(x) = \cosh(-x)$ symmetry holds globally without NaN or precision regressions.

CokieMiner:
Updated main pr description with all details, should be ready, some changes are from doing cargo fmt.

PR #265, opened 2026-05-18
Remove old tests

Noam2Stein:
Changes:

- Syncs the new `exp2` test with the improved old `exp2` tests.
- Removes the old tests.
- Removes the old test utilities (and refactors `tests/all_tests/t_usefulness.rs`).
- Moves `tests/all_tests/*` to `tests/*` and renames `t_*.rs` to `*.rs`.
- Removes the unnecessary `allow(...)` attributes in `tests/all_tests/main.rs` and fixes lints.

It seems all of the old tests are translated into the new format so removing the old tests should be fine.


Lokathor:
My recollection from many years ago is that the point of having an `tests/all_tests/main.rs` instead of separate test modules directly in the `tests/` folder is that the `tests/all_tests/main.rs` form makes all tests build in a *single* binary which runs all tests multi-threaded, and the `tests/` form makes a bunch of *separate* test binaries that build-link-run one after the other, and test multi-threading is only during each binary. So you end up with more build-link overhead by having separate test binaries, and each binary has to reset its thread worker pool and stuff when moving between binaries

@Noam2Stein would you mind just checking how long a clean build `cargo test` takes to run with and without this PR? It's probably not actually much, but I'd like to know if my memory is right on this one, and if it still makes a difference.

Noam2Stein:
I did think of that problem so i added `autotests = false` to the Cargo.toml then added a single manual `[[test]]` entry, that way all other files are ignored and only `tests/main.rs` produces a binary, but the directory is a bit cleaner.

These run `cargo clean` then `cargo test --timings`:

After flattening commit: compilation: 22.5s, running tests: 1.51s
Before flattening commit: compilation: 21.9s, running tests: 1.53s

Lokathor:
Ahhh, I see it there at the very end of the list of diffs. This is excellent then.

PR #266, opened 2026-05-18
Add inherit `simd_*` comparison functions and deprecate `Cmp*` traits

Noam2Stein:
The problem with comparison functions is that:

- Some types have inherit `simd_*` comparison function that don't support `rhs` being a scalar (unlike the trait methods).
- Other types don't have inherit functions and only implement the `Cmp*` traits (which must be imported by the user).

This fixes the problem by adding inherit `simd_*` functions with a generic parameter `Rhs` to support everything the trait methods support, and deprecating the traits, because if i understand correctly, the only reason they existed was to support `rhs` being a scalar?

Because there cannot be breaking changes, the new inherit functions use the deprecated traits as bounds, that way any use case in downstream code is not broken. Call sites don't need to import the traits themselves to use the new inherit functions.

Changes:

- Some `Cmp*` trait implementations previously called the old inherit functions. The implementations of old inherit functions were moved to the trait implementations.

- Added 6 `simd_*` inherit functions to each type via a macro. These new functions call the trait implementations to support all possible use cases and to not break downstream code.

- Deprecated the 6 `Cmp*` traits.

- Change tests to ensure everything works without importing the traits.


Lokathor:
this needs updating to the new test framework it seems, but other than that i like this direction

Lokathor:
Is this the last of the stuff you've got planned for the near future @Noam2Stein ? I can do a release if so.

Noam2Stein:
Yes that is it for now.

Noam2Stein:
Also, i think some changes are missing from the changelog (at least the float constants i remember).

Noam2Stein:
Hey @Lokathor its been a while, when you have the time could you make a release?

Lokathor:
released 1.5.0!

PR #268, opened 2026-05-20
Add more integer functionality

Noam2Stein:
Changes:

- Adds `clamp`
- Adds `is_positive`
- Improves `is_negative` for neon which has compare-to-zero intrinsics.
- Adds `signum`
- Adds `saturating_mul`
- Adds `Div` and `Rem`
- Adds `saturating_div`

Notes:

I tried to implement `saturating_mul` as efficiently as i can.

`clamp` doesn't panic for invalid inputs (`min > max`), but division does (`rhs == 0`) because all scalar operations do so there is no simple way to avoid it.

`Div`, `Rem` and `saturating_div` have zero hardware support and are all implemented with zero optimizations. I added documentation so that users know the functions are slow. I still think these should exist even though they have no optimizations, because if someone *does* need to use them, this is better than manually converting to an array.

Technically there are division optimizations like the ones used in `libdivide`, but i think those are only performant for a constant `rhs` (maybe a `div_const::<const RHS: ...>` function could be added in the future).

If you think division should not be added, those commits can be reversed.


Noam2Stein:
I cannot currently fix the bugs seen in the CI. I will fix them as soon as i can.

Noam2Stein:
@Lokathor I am tagging you in case the GitHub notifications were lost.

PR #270, opened 2026-06-09
Add `reduce_mul` for integers

### LINKED ISSUE #257: Feature request: `reduce_mul`
It would be useful to be able to calculate the product of the simd vector elements

It is already implemented in [std::simd](https://doc.rust-lang.org/std/simd/struct.Simd.html#method.reduce_product-4)

Noam2Stein:
Resolves #257.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # `wide` Changelog
 
+## Unreleased
+
+* Added `reduce_mul` for integers.
+
 ## 1.5.0
 
 * Added several functions and trait implementations that previously were only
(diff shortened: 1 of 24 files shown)


Noam2Stein:
It looks like the AVX512 CI fails for reasons unrelated to this PR.

SakiiCode:
Awesome, I think you can add "Resolves <span>#</span>257" to the description

Noam2Stein:
I didn't know you can do that! thanks

Lokathor:
The avx512 tests are failing for some sort of test environment download reason, and this doesn't change avx512 paths, so I'm gonna merge it.

PR #271, opened 2026-06-12
Add missing implementations of Mul<&Self>, MulAssign<Self> and MulAssign<&Self> for some int types

im-0:
Those were omitted for some reason despite having normal SIMD Mul for integer types.


Noam2Stein:
To make sure all types are covered its probably best to add a test for this.

im-0:
Are there existing tests for "trait X is implemented for Y" for other types? I looked briefly but was unable to find anything like this.

Noam2Stein:
You're right there actually aren't any tests for trait implementations like `AddAssign` etc. So they probably aren't needed for this PR specifically.

PR #272, opened 2026-06-18
impl Mul<u32x8> for u32

im-0:
This adds a missing trait implementation.

### DIFF
--- a/src/u32x8_.rs
+++ b/src/u32x8_.rs
@@ -194,6 +194,15 @@ impl Sub<u32x8> for u32 {
   }
 }
 
+impl Mul<u32x8> for u32 {
+  type Output = u32x8;
+
+  #[inline]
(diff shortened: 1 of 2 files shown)


Noam2Stein:
This implementation was supposed to be tested by `tests/simd.rs` `test_scalar_mul`. I see i mistakenly wrote `Simd::new(right) * left` instead of `left * Simd::new(right)` specifically in the integer path of that test, and that's how the missing implementation slipped through. Could you fix that? (if not ill make another PR)

(It looks like that's the only `test_scalar_*` test with that mistake, not including `shl`, `shr`, `bitand`, `bitor` and `bitxor` which have an explicit "no types have this implementation yet" note)

im-0:
> Could you fix that?

Done!

PR #277, opened 2026-06-28
fix, feat: Deprecate `blend` and split it into bitwise-blending and lanewise-blending

Noam2Stein:
# Reasoning

Currently `blend` doesn't have any specified behavior. I imagine it is used downstream as lanewise selection from a mask, and sometimes **incorrectly** as bitwise selection.

The current implementation sometimes works as bitwise selection and sometimes picks lanes based on the sign bit.

Since there are two use cases (lanewise from a mask and bitwise), and one of them is faster (lanewise) the function needs to be split into two functions.

# Changes

This deprecates `blend` and replaces it with two new functions `select` and `bitselect`. I chose these names but the functions can be renamed if there are better names.

`select` assumes each lane is either all zeros or all ones and is intended for masks. `bitselect` works as bitwise selection which is slower on x86 because there are no direct intrinsics for it.

`blend` now redirects to `select`, which has the original `blend` implementation without any changes. I chose to redirect `blend` to `select` and not `bitselect` because it already didn't work as bitwise selection.

If you think this change is too breaking perhaps there is a better solution, but this is a good baseline.

### DIFF
--- a/src/f32x16_.rs
+++ b/src/f32x16_.rs
@@ -421,19 +421,19 @@ impl f32x16 {
   /// [`select`]: Self::select
   #[inline]
   #[must_use]
-  pub fn bitselect(self, t: Self, f: Self) -> Self {
+  pub fn bitselect(self, if_one: Self, if_zero: Self) -> Self {
     pick! {
       if #[cfg(target_feature="avx512f")] {
         Self {
           avx512: bitor_m512(
-            bitand_m512(t.avx512, self.avx512),
-            bitandnot_m512(self.avx512, f.avx512),
+            bitand_m512(if_one.avx512, self.avx512),
+            bitandnot_m512(self.avx512, if_zero.avx512),
           ),
         }
       } else {
         Self {
-          a: self.a.bitselect(t.a, f.a),
-          b: self.b.bitselect(t.b, f.b),
+          a: self.a.bitselect(if_one.a, if_zero.a),
+          b: self.b.bitselect(if_one.b, if_zero.b),
         }
       }
     }
--- a/src/f32x16_.rs
+++ b/src/f32x16_.rs
@@ -709,10 +709,8 @@ impl f32x16 {
           cast(BOUNDS_LIMIT),
         ));
 
-        // `abs` keeps the original sign. `blend` cannot be used here because it
-        // doesn't work as an arbitrary bit-blend.
-        let bounds_mask = bounds_mask.abs();
-        result_abs & bounds_mask | self & !bounds_mask
+        // `abs` keeps the original sign.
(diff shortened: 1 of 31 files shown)


src/f32x16_.rs
```diff
@@ -410,16 +410,54 @@ impl f32x16 {
 
   simd_comparison_fns!();
 
+  /// Bitwise selection.
+  ///
+  /// For each bit this returns `t` where `self` is `1` and `f` where `self` is
+  /// `0`.
```
RagnarGrootKoerkamp: It's not so clear here that `t` and `f` indicate only the single corresponding bit of the input.

RagnarGrootKoerkamp (commented):
portable-simd also calls this `select`, so that seems fine.

But `bitselect` is probably easily misunderstood as the "selecting with bitmask" described [here](https://doc.rust-lang.org/std/simd/trait.Select.html#selecting-with-a-bitmask), where the control mask is a `u64` and each bit controls a lane.

Which types had the wrong implementation? I suppose this would technically be a breaking change.

Noam2Stein:
Most types had a wrong implementation for x86 when at least `sse4.1` is enabled.

This technically isn't a breaking change since the existing `blend` function is unchanged, but now has documentation that explains the problem. We could instead redirect it to `bitselect`, which would be slightly slower but would fix the existing `blend` bug, if you consider it a bug.

Personally i like how `bitselect` mirrors `select` (for example unlike `bitblend`) so i am not sure if there is a better name.

src/f32x16_.rs
Noam2Stein: Ill improve the documentation and update the parameter names later.

RagnarGrootKoerkamp:
Ahhh, I see my confusion. You wrote:

> The current implementation sometimes works as bitwise selection and sometimes picks lanes based on the sign bit.

Now that it is documented that the input must be a proper mask, these become equivalent for `select`, and so no code changes are actually needed to "fix the bug".

I agree it should just forward to `select`.

`bitselect` is probably fine as long as the docs are precise enough.

Lokathor:
we could use `bitwise_select`, that's maybe clearer that you're selecting *per bit*.

but `bitselect` is also "weird enough" that people will probably pause and check what the docs say it does, so `bitselect` is also fine.

Noam2Stein:
The upsides I see for `bitselect` are:

- Its concise and less verbose
- Its consistent with `bitand`, `bitor`, `bitxor`
- The name is also used in wasm intrinsics and apparently OpenGL

More alternatives are `select_bits` and `bit_select`, but I think based on Rust's name conventions `bitselect` wins.

PR #280, opened 2026-06-29
fix: Edge case in `powf`

Noam2Stein:
This fixes a bug in `powf` which was marked with this TODO:

```
// TODO: fix `powf` which currently breaks when, possibly among other cases,
// `self` is negative and `n` is an odd number. These inputs lead to an
// incorrect result:
//
// simd_chunks!(
//  [
//    1.2, 2.0, 3.0, 1.5, 9.2, 6.1, 2.5, 5.3, -4.5, -5.1, -5.2, -5.3, -3.0,
//    -3.1, -3.0, -4.0, 5.1,
//  ],
//  [
//    0.1, 0.5, 1.0, 2.718282, 3.0, 4.0, 2.5, -1.0, 1.4, 2.0, 1.0, 3.0, 0.1,
//    2.7, 4.0, -3.0, 29.0,
//  ],
// )
```

# Explanation

The bug did occur specifically when `x` is negative and `n` is an odd integer. The implementation checks for this case and is supposed to flip the sign of the result if it occurs by shifting the least significant bit of `round_int` into the sign bit. The bug was caused by using `round_float` instead of `cast` to convert the shifted sign bit value into a float.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,8 @@
 
 * Fixed bugs in the fallback paths of `any`, `all`, `none` and `fast_clamp`.
 
+* Fixed bug in `powf`.
+
 ## 1.5.0
 
 * Added several functions and trait implementations that previously were only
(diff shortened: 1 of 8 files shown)


Lokathor:
after merging your other PRs, this one and https://github.com/Lokathor/wide/pull/277 ended up having conflicts

Issue #65, opened 2020-09-19
Add documentation


RazrFalcon:
It would be great to have a documentation, especially Intel-like: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#expand=3333,100,100&text=_mm512_add_epi32

```
FOR j := 0 to 15
	i := j*32
	dst[i+31:i] := a[i+31:i] + b[i+31:i]
ENDFOR
dst[MAX:512] := 0
```

Lokathor:
Yeah, docs would be good.

Most functions can probably just be a note like "lanewise add" or "lanewise sqrt" or similar.

I want to avoid excessive rustdoc example code because that makes the test take far longer to run through.

I'd accept PRs with docs, but probably won't end up doing this myself because the Portable SIMD project group has officially begun, so most of my energy will go there. Ideally what comes out of that project will totally supercede this crate without any loss of ergonomics.

valflrt:
I think documentation is crucial for functions such as blend which are not always obvious for people new to simd.

Noam2Stein:
I am willing to add documentation, but first i have two questions:

1. They probably shouldn't be Intel style (as in the top comment) and should instead explain semantics and guarantees similar to the standard library documentation for `f32`/`f64`. Is this agreed upon?

2. Currently all documentation is duplicated for each type that has a given function. Is there some way to avoid this duplication (maybe there is a good approach with macros) or should i add it duplicated?

Noam2Stein:
This macro approach seems quite maintainable and flexible:

```rust
pub struct f32x4([f32; 4]);

macro_rules! simd_api {
    (
        $select:item

        $transpose:item
    ) => {
        /// Lanewise selection.
        ///
        /// Bla bla bla.
        #[must_use]
        $select

        /// Docs.
        #[must_use]
        $transpose
    };
}

impl f32x4 {
    simd_api! {
        #[inline]
        pub fn select(self, if_true: Self, if_false: Self) -> Self {
            todo!()
        }

        #[inline]
        pub fn transpose(array: [Self; 4]) -> [Self; 4] {
            todo!()
        }
    }
}
```

With this approach you can write documentation once for all types, but can still opt out for an individual function if its documentation needs to be different for each type. I don't think this would impact compile times but ill check if its implemented.

Lokathor:
I like the macro approach, and i like the "real explanation" approach rather than giving pseudo-code and forcing the user to read it and try to understand what happened.

[...]

PR #284, opened 2026-07-02
refactor: Move the library's functionality into consistent macros

Noam2Stein:
This PR implements the macro idea mentioned [here](https://github.com/Lokathor/wide/issues/65#issuecomment-4829142436). This should make it a lot easier to later add consistent documentation for all types, and to maintain the crate in general.

Benefits are:

- Documentation for a function can be defined once for all types while each type has a unique implementation.
- Functions that have the same implementation for all types can be added without touching each individual type.
- This makes it a lot easier to keep all types consistent with each other.

Downsides are:

- If a macro call (for example for `impl_simd_float`) gets the order of functions incorrect accidentally, documentation can get swapped between functions and most likely there won't be a compile error.

# Large git diff

Note that this PR has an extremely large git diff, because it moves and reorders most of the crate's functions so that they fit cleanly in the new macros.

This PR is separated into many small commits that slowly move functionality to the new macros, which will hopefully make it a bit easier to reliably review this PR. Tell me if there is a way to make this PR easier to review.

# Documentation

This PR generally doesn't add new documentation. Some functions that had documentation only for specific types now have documentation for all types.

The functions `to_bitmask`, `any`, `all` and `none` only had documentation for `i64xN` types, and the documentation promises a guarantee that I am not sure is a good idea (the guarantee is that the most significant bit is used as the flag of the lane).

Because of this, documentation is kept only for `i64xN` types, but `i64xN::none` now looses its documentation because `none` is implemented automatically and identically for all types.

# Problem: Compile times

The new macros don't seem to hurt compile times, but an indirect change in this PR does.

The commit "Move `-!+-*/%<<>>&|^` operators (for both floats and integers)" adds all possible variants of binary operators (simd x simd, simd x scalar, scalar x simd, reference variants). Previously these variant implementations existed inconsistently per operation and type.

Adding all implementations heavily hurts compile times, specifically from `2.5s` to `4.0s` on my machine.

I am not sure whats the correct solution to this. I could keep only the implementations that already existed previously, but that requires a long scan over the source code of the crate to find out exactly which variants existed, so I will first wait for a decision for what should be done before spending time scanning the library.

# Small notes

The previous implementation of the `UpperExp` formatting trait for floats was incorrect and is now fixed. Values were incorrectly formatted as integers instead of as floats.


Noam2Stein:
These are extra notes that I forgot about while writing the top comment:

Another downside to this macro system is that rustfmt fails to parse the macro calls, this is not a big issue since the `pick` macro already disables rustfmt anyway, but its worth noting.

Since most of this PR is copy pasting code around, its important to note exactly where there are changes where code isn't being exactly copy pasted. The implementations for `simd << scalar` and `simd >> scalar` were previously generated by macros directly for each possible `rhs` type (all primitive integers), but now only the `u32` version exists directly, and all other implementations redirect to that implementation (which could slightly impact code generation). Small but worth noting. The new implementation also adds support for `rhs` being `usize` and `isize`.

Lokathor:
i think even with the compile time increase, this is a good change to have full consistency

Lokathor:
should i merge now or is there more to do?

Noam2Stein:
Yes this can be merged, as long as the compile time increase is fine.

Issue #285, opened 2026-07-02
Typos: "inherit" instead of "inherent"


RagnarGrootKoerkamp:
The deprecation message for `cmp_eq` (and the changelog) has "inherit" instead of "inherent".
Would also be nice to mention there that the new method is called `simd_eq` instead of `cmp_eq`.

Noam2Stein:
I think this is also resolved by the merged PR?

PR #286, opened 2026-07-03
docs: Add documentation for everything

Noam2Stein:
This solves #65 and #285 (since it is a documentation issue).

This adds documentation for the entire crate. Its now possible to lint against `missing_docs` (should we lint against that? if yes, `warn`, `deny` or `forbid`?).

In addition to making sure the new documentation explains things well, I would also like feedback about the specific guarantees the documentation of each function makes, since getting those wrong then reverting them later would be a breaking change.

# Issue with guarantees of certain functions

The functions `to_bitmask`, `any`, `all` and `none` previously had no documentation for most types, but did have documentation specifically for `i64xN` types.

By intuition I would expect these functions to assume `self` is a mask, and to not guarantee anything about what bits are used, but the existing documentation for `i64xN` promises these functions use the most significant bit of each element as the bit deciding whether the element is considered `true` or `false`.

What should be done with these? We could either add the guarantee for all types or revert it for `i64xN` (which would be a breaking change). Right now these functions are marked with TODOs.


Noam2Stein:
Now that I rebased this branch on top of `main`, this PR is ready for reviews.

Noam2Stein:
Another note: It may also be a good idea to review this from the perspective of the generated `cargo doc` output.

Lokathor:
unfortunately, it looks like i caused a bunch of conflicts when i merged the other PRs.

Noam2Stein:
Thats fine. Other than conflicts, remember before merging this PR there is the problem with the four functions listed above, the question of whether we should lint against `missing_docs` (I'd prefer so so that future contributions will have a higher standard in terms of documentation), and nitpicks about added documentation if there are any.

Lokathor:
Definitely lint against missing_docs!

Lokathor:
Alright @Noam2Stein if you fix this merge conflict up too I promise to merge this one before anything else.

Noam2Stein:
There are a couple more things I forgot to add to documentation before merging, and what should be done about the breaking change problem in the top comment?

Noam2Stein:
I fixed the merge conflicts and added everything I forgot about earlier. All thats left now is deciding what to do with the breaking change.

Lokathor:
I64xN need to have the change, and technically it's a break one could say, but also it's a big one could say. If it makes the type more like every other type in the crate, then we call it a bug fix and do it.

Noam2Stein:
As far as I'm concerned this can be merged now.

PR #289, opened 2026-07-10
feat, refactor: Add more integer multiplication functions

Noam2Stein:
Changes:

- Renames `mul_widen` to `widening_mul`, deprecates the original functions and adds the new one to remaining types (excluding types that don't have a wider variant).
- Adds `mul_keep_high` to remaining integers (and moves it to the new macros).
- Adds `mul_keep_low_high`.
- Recycles `mul_keep_low_high` for `saturating_mul` and `overflowing_mul`

Currently std's nightly `widening_mul` function is actually equivalent to the added `mul_keep_low_high` function because it returns a tuple `(low, high)`, but based on the tracking issue it looks like its going to change to returning a wider integer type instead, like the `widening_mul` function added in this PR.

`mul_keep_low_high` returns a tuple of two SIMD vectors, one for the low bits and one for the high bits. I am not sure if its current name is great, it could be `mul_low_high` instead, but then `mul_keep_high` may need to be renamed too.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* Added `reduce_mul` for integers.
+* Added integer functions `reduce_mul` and `mul_keep_low_high`.
 
 * Added overflowing arithmetic for integers.
 * Added float function `round_ties_even`.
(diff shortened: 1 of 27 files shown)


Lokathor:
`mul_keep_low_high` is sufficently named. it's a silly name, but it says what it does.

PR #293, opened 2026-07-11
Support u8x64 and i8x64 types

RagnarGrootKoerkamp:
Just had claude add these. 

The diff between these and `u16x32` and `i16x32` is:
- `u8` multiplication always falls back to the two 256-bit halves.
- `shl` and `shr` work on the two halves like `u8` does for other widths. Would be nicer to replace this by a `u16x32` shift and mask though, at least for shifts by a constant.
- (I kept the somewhat broken comments from #264 cc @Noam2Stein.)
- I did not add the `dot` function of `i16x32` because` mul_i8_horizontal_add_m512i` does not exist in `safe_arch`.
- I did not add the `swizzle_half{_relaxed}` functions that eg `u8x32` has.

### DIFF
--- a/src/f32x16_.rs
+++ b/src/f32x16_.rs
@@ -8,7 +8,7 @@ pick! {
   } else {
     #[derive(Default, Clone, Copy, PartialEq)]
     #[repr(C, align(64))]
-    pub struct f32x16 { pub(crate) a : f32x8, pub(crate) b : f32x8 }
+    pub struct f32x16 { pub(crate) a: f32x8, pub(crate) b: f32x8 }
   }
 }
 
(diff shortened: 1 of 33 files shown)


Noam2Stein:
This also means `widening_mul` can be added for `u/i8x32` (in this PR or a separate one)

Edit: What I said is incorrect, `u/i8x64` has nothing to do with `widening_mul` for 8x32.

Lokathor:
looks like the CI failed because of some missing methods. there has been a lot of internal shuffling recently so you might just need a small update.

RagnarGrootKoerkamp:
Sorry I missed the messages here. See #318 for a new attempt.

PR #296, opened 2026-07-15
refactor: Rename `powf` functions

Noam2Stein:
Currently, the powf function that that is SIMD x SIMD has a different name for each type (e.g., `f32x4::pow_f32x4`, `f64x4::pow_f64x4`), and the function name `powf` is used for the SIMD x scalar version, which internally just calls the SIMD x SIMD version with a `splat`.

When using macros for each SIMD type, its quite annoying that the SIMD x SIMD function has a different name for each type. Instead there should be one consistent name. Also since the SIMD x scalar version just calls the SIMD x SIMD version (meaning it doesn't have some sort of optimization) there is no real reason for it to exist.

Changes:

- Renames the SIMD x SIMD function to `powf_simd` for all float types. (without the `f` this could be confused with `powi`, so its important to keep it) (cannot be named `powf` because the name is taken)
- Deprecates the original functions.
- Deprecates the scalar function `powf` and recommends using `splat` at the call site instead (if in the future an actual optimization is written specifically for SIMD x scalar, a `powf_scalar` function can be added)
- Uses documentation copied from #286 next to naming explanations.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* Renamed float function `pow_{simd-type-name}` to `powf_simd` and deprecated `powf`.
 * Added conversions between `wide` types and native intrinsics SIMD types.
 * Added `reduce_mul` for integers.
 * Added integer functions `reduce_mul` and `mul_keep_low_high`.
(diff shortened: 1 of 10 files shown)


Lokathor:
The best part of all these PRs is that you always put notes in the changelog for me, I barely have to do a thing to make a release!

Noam2Stein:
Thats great! One of my goals as a contributor is to make the job of the maintainer as easy as I can.

PR #297, opened 2026-07-16
feat, build-perf: Add remaining shifts and reduce operator compile time

Noam2Stein:
This is a small change I thought of while writing another PR.

Changes:

- Adds support for shifting a signed SIMD vector by the equivalent unsigned SIMD vector, and shifting an unsigned SIMD vector by the equivalent signed SIMD vector. (and the same thing for scalar and reference variants)
- Replaces the manual "all combinations" approach previously used to implement assign ops and reference implementations for more generic implementations, reducing compile times from `3.9s` to `3.5s` on my machine.
- Recycles function implementations of unsigned types for signed types where possible (meaning signed functions call unsigned functions where the logic is the same). This reduces compile times from `3.5s` to `3.4s` on my machine.

I tried doing the build optimization for standard binary operators as well, but for them it actually hurts compile times. I didn't recycle integer function implementations for floats (like signed recycles unsigned) because they still use different instructions that go into different CPU execution units.


Lokathor:
oops, caused a merge conflict, and i can't tell which way to resolve it because it's deep in a macro

PR #299, opened 2026-07-17
feat: `cast_unsigned`, `cast_signed`, `to_bits` and `from_bits`

Noam2Stein:
These functions are easier to use than fully-typed bytemuck casts, and are consistent with `std`. Added functions are `const` because they are simple bitwise casts.

Since these are so simple I don't think there is a point to adding tests.


Lokathor:
conflict here too

PR #305, opened 2026-08-06
feat: Add add_mul_lo/hi, swizzle2, and unpack_lo/hi for unsigned integers

runesmalediction:
This adds `add_mul_lo`, `add_mul_hi`, `unpack_lo`,  `unpack_hi` and `swizzle2` for `u32xN` and `u64xN` types.

`add_mul_lo` and `add_mul_hi` compute `self + ((a * b) mod 2^W)` and `self + ((a * b) >> W)`, reading only the low `W` bits of each lane of `a` and `b`.  This is the operation AVX-512-IFMA exposes as `vpmadd52luq`/`vpmadd52huq`, and the `u64xN` versions lower to those directly when `W == 52`. Narrower widths (`W <= 32` for `u64xN`, `W <= 16` for `u32xN`) use a widening multiply, since the whole product still fits in a lane.  Wider ones fall back to scalar.

`unpack_lo` and `unpack_hi` interleave the low or high half of two vectors one element at a time, so for `u32x8` `unpack_lo` returns `[self[0], b[0], self[1], b[1], self[2], b[2], self[3], b[3]]`. I had first used the native `unpack` instructions that act within each 128-bit lane, but this is a different operation. The current matches what is in `i64x4` and `f64x4`.

`swizzle2` gathers lanes from the concatenation of two vectors. 

Each of the new functions has an accelerated path behind `pick!` and a scalar fallback.

I needed this for my own project, so I had Claude add them as I am unfamiliar with most SIMD intrinsics except for NEON. I did try to put in some effort to go over everything myself and ensure the code quality is acceptable.

A couple of points that might prevent merging:
- Some intrinsics in `unsafe { ... }` blocks, should those be added to `safe_arch` first?
- I left some helper functions in `lib.rs`, but these might fit better in their own module.
- The W == 52 special case only exists because that is the width AVX-512-IFMA is fixed at, but might be a bit obscure to users (I do use it myself in my own project).

I am more than happy to make any changes (without the use of LLMs if preferred :smile: ), just let me know.

EDIT: `swizzle2` is dropped from this PR, #317 covers it with a different design. The unsafe intrinsics are now safe_arch 1.2.0 wrappers, so only the NEON paths stay unsafe.


Noam2Stein:
I'm not familiar with the use cases of `add_mul_lo/hi`, but is there a reason these are not named `mul_add_*` instead `add_mul_*`? (for consistency with the float functions) and the order being `((self * a) mod W) + b`

runesmalediction:
> I'm not familiar with the use cases of `add_mul_lo/hi`, but is there a reason these are not named `mul_add_*` instead `add_mul_*`? (for consistency with the float functions) and the order being `((self * a) mod W) + b`

It's the difference in accumulator position between the two. `add_mul_lo/hi` does `self + ((a * b) mod 2^W)`, so `self` is the accumulator, while `mul_add` does `(self * a) + b`, where `self` is a multiplicand. The accumulator ordering mirrors the instruction: the IFMA path is a direct call to `_mm256_madd52lo_epu64(self, a, b)`, and `vpmadd52luq` accumulates into its destination.

I think this is correctly marked in the repo, but of course I'm more than happy to changed it around.

[curve25519-dalek's IFMA backend](https://github.com/dalek-cryptography/curve25519-dalek/blob/main/curve25519-dalek/src/backend/vector/ifma/field.rs#L31) is an example of a use case of add_mul_*.

Noam2Stein:
About `swizzle2`, notice how you added it for 32-bit and 64-bit types, but currently `swizzle` only exists for 8-bit types.

I made a PR adding `swizzle` for all types (though renamed), so lets see if that gets approved, then could you add `swizzle2` (though maybe renamed too) for all types (including 8-bit and 16-bit)?

About performance, when there isn't hardware support (which without avx512 it looks like there isn't any for 16-bit 32-bit and 64-bit swizzling) you can emulate the operation by turning your 16/32/64-bit indices into byte indices that can be used for 8-bit swizzle intrinsics (see the implementations of `swizzle_dyn` in #311), which should be a lot faster than falling back to `from_fn(|i| ...)` (another tip, on `aarch64` you get `vqtbl2q_u8`, which performs `swizzle2` natively for 8-bit elements, and that can also be used to emulate 16/32/64-bit).

runesmalediction:
Thanks for the tip on the byte-wise swizzle fallback, it's implemented now: byte shuffle on SSSE3/simd128, vqtbl2q_u8 on aarch64, and vpermd on AVX2. I'll rename/implement swizzle2 following #311 as soon as it merges.

Noam2Stein (commented):
After thinking about it, it may be a bad idea exposing `swizzle` and `swizzle_2` as separate functions, since later `swizzle_3` and `swizzle_4` could also be added (`neon` supports them natively), and that would result in a huge number of functions. Instead, I have a simpler design for swizzle/shuffle methods, and ill make it into a PR, so lets not merge this part of this PR (I also closed #311).

About `add_mul_*`, I am not familiar enough with the operation to decide if its appropriate to expose the const generic API in the PR, so maybe its better if @Lokathor or someone else decides on that part.

The `unpack_lo/hi` functions are great (if you want, you could split those into their own PR so that they can get merged before `add_mul_*`).

Lokathor:
I only had a moment to check on the PRs this evening, but I can say that 

1) I don't mind more private helpers in `lib.rs`, that's totally fine. i love the const asserts too, very cool being on a newer compiler compared to when i wrote most of this long ago.
2) we should absolutely add intrinsic wrappers to `safe_arch` and then have `wide` make safe calls as often as possible. If you have the time for that, please do that.

runesmalediction:
Opened Lokathor/safe_arch#127 with the wrappers, 13 in total plus a small gate fix. Once that's released I'll bump the dependency here and swap the raw intrinsics over to safe calls. The NEON paths will have to stay `unsafe` though, since `safe_arch` is x86/x64 only.

Lokathor:
released safe_arch-1.2.0, so now you can update the PR to use it in all the new places. ~~don't forget to update the dependency line in `Cargo.toml` too, or people that build with the minimum versions flag will have a bad time.~~ EDIT: oh you already mentioned that, right! EDIT2: yeah i'd love to have safe_arch do neon some day, but it's a whole lot of work for not much practical gain now that we have target_feature_1.1

Noam2Stein:
Also, remember to remove the `swizzle2` methods, since #317 now adds those with a different design (and thanks for the nice `to_byte_indices` helper functions, I reused those, they are a lot cleaner than my original code).

runesmalediction:
`swizzle2` is removed, #317 covers it with a better design.

`safe_arch` is bumped to 1.2.0 and the new x86 paths go through the wrappers now, so the only unsafe left is the four NEON `vzip1q`/`vzip2q` calls in `unpack_lo/hi`. 

Also merged latest main.

Noam2Stein:
It seems this PR is complete right?

The following comments are just my personal opinion on the `add_mul_*` API, so @Lokathor if you think *it is* a good idea to have it then merge this.

I'm not too familiar with the use cases, but personally I think the `add_mul_*` API is too niche to be added to `wide` because 1. I don't think its a fundamental operation 2. `std` integers don't have it 3. It seems the only case where there is performance benefit is for the `W=52` case.

I too have a case where I need an operation I think is too niche for `wide` (specifically a cross platform deterministic `f32x4::reduce_add` performing `(x+y)+(z+w)`). I think that in such cases its best to use intrinsics directly, and thats why I added #290.

Lokathor:
It's... Fine enough. People not using the methods aren't really hurt by having them in here. I'll go ahead and merge this.

PR #307, opened 2026-08-07
Fix 306

### LINKED ISSUE #306: wide 1.6.0 fails to compile when AVX-512F is enabled without AVX-512DQ
## Description

`wide 1.6.0` fails to compile when `avx512f` is enabled but `avx512dq` is not.

This is a valid Rust target-feature configuration. Rust models the AVX-512 subsets separately: `avx512bw` and `avx512vl` imply `avx512f`, but neither implies `avx512dq`.

Some wide vector types select their storage representation based on `avx512f`, while some methods select their implementation based on `avx512dq`. As a result, under `avx512f && !avx512dq`, the types contain an `avx512` field, but the fallback implementations try to access the non-AVX-512 `a` and `b` fields.

The failure occurs even when no API from `wide` is explicitly used.

## Minimal reproduction

`Cargo.toml`:

```toml
[package]
name = "wide-avx512-repro"
version = "0.1.0"
edition = "2024"

[dependencies]
wide = "=1.6.0"
```

`src/main.rs`:

```rust
fn main() {}
```

Run:

```powershell
$env:RUSTFLAGS = '-C target-cpu=x86-64 -C target-feature=+avx512f,+avx512bw,+avx512vl'
cargo check
```

The requested configuration intentionally does not enable `avx512dq`.

It can be confirmed with:

```powershell
rustc --print cfg `
  -C target-cpu=x86-64 `
  -C target-feature=+avx512f,+avx512bw,+avx512vl
```

The output includes:

```text
target_feature="avx512bw"
target_feature="avx512f"
target_feature="avx512vl"
```

and does not include:

```text
target_feature="avx512dq"
```

## Actual result

Compilation fails with 12 `E0609` errors. Representative errors:

```text
error[E0609]: no field `a` on type `f64x8_::f64x8`
  --> wide-1.6.0/src/f64x8_.rs:658:16
   |
658 |           self.a.round_int(),
   |                ^ unknown field
   |
   = note: available field is: `avx512`
```

```text
error[E0609]: no field `a` on type `i32x16_::i32x16`
  --> wide-1.6.0/src/i32x16_.rs:160:14
   |
160 |         self.a.to_bitmask() | (self.b.to_bitmask() << 8)
   |              ^ unknown field
   |
   = note: available field is: `avx512`
```

The affected methods include at least:

* `f64x8::round_int`
* `f64x8::fast_round_int`
* `f64x8::trunc_int`
* `f64x8::fast_trunc_int`
* `i32x16::to_bitmask`
* `i64x8::to_bitmask`

## Suspected cause

For example, `f64x8` selects its representation using `avx512f`:

```rust
pick! {
  if #[cfg(target_feature="avx512f")] {
    pub struct f64x8 {
      pub(crate) avx512: m512d,
    }
  } else {
    pub struct f64x8 {
      pub(crate) a: f64x4,
      pub(crate) b: f64x4,
    }
  }
}
```

However, `round_int` selects its implementation using `avx512dq`:

```rust
pick! {
  if #[cfg(target_feature="avx512dq")] {
    // Accesses self.avx512.
  } else {
    // Accesses self.a and self.b.
  }
}
```

Therefore, when `avx512f` is enabled and `avx512dq` is disabled, the fallback branch accesses fields that do not exist in the selected representation.

## Expected result

The crate should compile for a target configuration that enables AVX-512F without AVX-512DQ.

Operations that require AVX-512DQ could use a fallback compatible with the AVX-512F storage representation, or the representation and implementation `cfg` conditions could otherwise be made consistent.

## Workaround

Explicitly enabling `avx512dq` makes the crate compile:

```powershell
$env:RUSTFLAGS = '-C target-cpu=x86-64 -C target-feature=+avx512f,+avx512bw,+avx512dq,+avx512vl'
cargo check
```

## Environment

```text
wide 1.6.0
rustc 1.97.1 (8bab26f4f 2026-07-14)
host: x86_64-pc-windows-msvc
LLVM version: 22.1.6
```

Lokathor:
* Closes https://github.com/Lokathor/wide/issues/306


Lokathor:
@Noam2Stein i can't currently formally assign you as a reviewer (though i sent you a repository collaborator invite), but this is all correct enough right?

Noam2Stein:
I am honored that you made me a collaborator! Yes this looks correct, and the CI reproduces the exact setup from the issue so the problem should be fixed.

Noam2Stein (approved):

Noam2Stein:
Even though I have write access, I think its safer if I only review and approve changes, but wait for you to merge them (whatever workflow you think is best here)

PR #311, opened 2026-08-08
feat: Add dynamic swizzle functions for all types

Noam2Stein:
Edit: I actually have more changes to make, so don't merge yet.

This PR adds two functions to all SIMD types:

- `swizzle_dyn`: dynamic swizzle from one input vector. Does not specify the result if indices are out of bounds, except for 8-bit element types which make additional guarantees.

- `zeroing_swizzle_dyn`: dynamic swizzle from one input vector. If an index is out of bounds, zero is returned.

These exist for *all* SIMD types, not just `i/u8` ones. Even though these functions mostly don't have hardware support for non-u8 types, they still are more optimal than the user manually indexing arrays, so I think they *should* exist (most implementations use mul+add+8bitshuffle to emulate it).

Unlike the old `swizzle` functions, which for signed integer types take *signed* integers as indices, these new functions always take unsigned integers as indices (for floats, signed integers, unsigned integers). This is one of the reasons I didn't reuse the original function names.

The original `swizzle` and `swizzle_relaxed` functions that exist for 8-bit element types are deprecated. The precise guarantees `swizzle_relaxed` made for each type are kept in `swizzle_dyn` for those types.

Before merging:

- Are there better names for these functions? (the logic behind `swizzle_dyn` is that the `dyn` should be explicit, since dynamic swizzle is more niche than constant swizzle, and the logic behind `zeroing_swizzle_dyn` is that in the future `wrapping_swizzle_dyn` can also be added, since I *think* that on avx512 that is the native behavior).

- (edit: done) I want to make some more optimizations before merge

- (edit: done) I want to make sure I didn't miss any existing `safe_arch` functions, and make sure all implementations are correct

In the future, these can be extended with `swizzle_2_dyn` (like in the open PR #305), and some form of what is currently `swizzle_half` (I am not sure how that should be named).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,9 @@
 * Fixed the `signum` documentation: the docs previously misstated the grammar
   and the set of inputs that map to `1.0`/`-1.0`. The behavior itself was
   already correct and matches [`f32::signum`].
+* Added dynamic swizzling functions `swizzle_dyn` and `zeroing_swizzle_dyn` for
+  all types, and deprecated the old 8-bit element functions `swizzle` and
+  `swizzle_relaxed`.
 
 ## 1.6.1
 
--- a/src/f32x16_.rs
+++ b/src/f32x16_.rs
@@ -191,6 +191,16 @@ impl_simd! {
     }
   }
 
+  #[inline]
+  pub fn swizzle_dyn(self, idxs: u32x16) -> Self {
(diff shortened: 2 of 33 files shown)


Noam2Stein:
I decided to close this, since I thought of a simpler shuffling API than this, so ill make a new PR.

PR #313, opened 2026-08-09
feat: add accelerated u8x32 shifts (closes #208)

### LINKED ISSUE #208: Missing `u8x32` shifts
For some reason, AVX2 does not have shifts for `u8x32` (and probably also not for smaller `u8xL`), and so neither does `wide` currently.
At least shifts by a single scalar constant `s` can be implemented instead by casting to `u16x16`, doing the shift there, and then masking out the high `s` bits of each `u8`.

Not sure what the status is around non-scalar shifts.

A note here is that `portable-simd` forwards all these operations to intrinsics directly, so looking at its source code for inspiration doesn't really work.

bravo1goingdark:
Closes #208.

**TL;DR:** `u8x32` now has fast shift paths where the hardware allows it —
~5x faster per-lane shifts on AVX-512 and 2-4x faster scalar `shr` on
AVX2/AVX-512, by shifting in `u16` lanes and truncating back to `u8`. The
scalar `shl` variants stay on the fallback because the manual AVX2 version
measured slower than auto-vectorization. Everything else behaves as before.

## Approach

x86 has no byte-level shift instruction, so this follows the suggestion in the
issue: widen to `u16`, shift in the wider lanes, then truncate back to `u8`
(masking to the low 8 bits before the saturating pack, since `shl` results can
exceed `u8::MAX`).

- vector-count variants (`shl`, `shr`, `unbounded_shl`, `unbounded_shr`) use
  `avx512bw`+`avx512vl`, shifting each lane by its own count
- scalar-count `shr` variants use `avx2`
- scalar-count `shl` variants intentionally keep the fallback (measured
  slower than auto-vectorization, ~2.95ns vs ~2.59ns)
- non-x86 targets behave exactly as before

Note on the unbounded variants: the count is clamped to 8 rather than 7,
because shifting a byte by 8 in `u16` and keeping the low byte gives 0, which
is what `rhs >= 8` is supposed to produce.

## Bugs fixed along the way

Early versions had bugs that only appeared with AVX2/AVX-512 enabled:

- scalar shifts returned all zeros (shift count was passed as a splat, but
  `_mm256_sll_epi16` reads it from the low 64 bits of the register)
- `shl` of values overflowing a byte were wrong (saturating pack clamps where
  truncation was needed)
- the AVX-512 path scrambled lanes (re-widened already-widened data, and
  extracted the wrong 256-bit half)
- `unbounded` shifts with `rhs >= 8` were wrong (clamped to 7 instead of 8)

## Benchmark

Single op in a tight loop on my machine (11 trials, min):

| op | plain | AVX2 | AVX-512 |
| --- | --- | --- | --- |
| vector-count shifts | ~11.5 ns | n/a | **~2.2 ns** |
| `shr` scalar | ~3.2 ns | ~1.5 ns | **~0.8 ns** |
| `unbounded_shr` scalar | ~3.2 ns | ~0.9 ns | **~0.8 ns** |

Roughly 5x faster vector shifts on AVX-512, 2-4x faster scalar `shr` on
AVX2/AVX-512. On AVX2-only machines the vector-count ops stay at fallback
speed — AVX2 simply has no byte-level variable shift.

## Testing

176/176 tests pass on plain, `+avx2`, and `target-cpu=native` builds; clippy
clean. Only `src/u8x32_.rs` and `changelog.md` touched.


src/u8x32_.rs
```diff
@@ -248,12 +248,15 @@ impl_simd_uint! {
...
   #[inline]
   fn shl(self, rhs: Self) -> Self::Output {
-    // For x86, this technically can be done explicitly by converting to `u16`
-    // or `u32` then converting back after multiplication, but that may not
-    // actually be faster than auto-vectorization.
-    let [self_a, self_b]: [u8x16; 2] = cast(self);
-    let [rhs_a, rhs_b]: [u8x16; 2] = cast(rhs);
-    cast([self_a << rhs_a, self_b << rhs_b])
+    pick! {
+      if #[cfg(all(target_feature="avx512bw", target_feature="avx512vl"))] {
```
RagnarGrootKoerkamp, line 252: I really want a SIMD version specifically for AVX2 though...

src/u8x32_.rs
```diff
@@ -267,21 +270,29 @@ impl_simd_uint! {
...
 
   #[inline]
   fn shr(self, rhs: u32) -> Self::Output {
-    // For x86, this technically can be done explicitly by converting
-    // to `u16` or `u32` then converting back after multiplication, but that
-    // may not actually be faster than auto-vectorization.
-    let [self_a, self_b]: [u8x16; 2] = cast(self);
-    cast([self_a >> rhs, self_b >> rhs])
+    pick! {
+      if #[cfg(target_feature="avx2")] {
```
RagnarGrootKoerkamp, line 287: why no specific AVX512 version here if that would be faster?

src/u8x32_.rs
```diff
@@ -533,3 +558,56 @@ impl u8x32 {
...
+  // work on the two 128-bit halves separately (that's all AVX2 has).
+  #[inline]
+  fn shift_all_u16(self, count: m128i, right: bool) -> Self {
+    let low = convert_to_i16_m256i_from_u8_m128i(extract_m128i_m256i::<0>(self.avx));
+    let high = convert_to_i16_m256i_from_u8_m128i(extract_m128i_m256i::<1>(self.avx));
+    let low = if right { shr_all_u16_m256i(low, count) } else { shl_all_u16_m256i(low, count) };
+    let high = if right { shr_all_u16_m256i(high, count) } else { shl_all_u16_m256i(high, count) };
+    let low = bitand_m256i(low, set_splat_i16_m256i(0xFF));
+    let high = bitand_m256i(high, set_splat_i16_m256i(0xFF));
+    Self { avx: Self::pack_u16_halves(low, high) }
```
RagnarGrootKoerkamp, line 598: This pretty much looks super complicated for what we're trying to do here?

src/u8x32_.rs
bravo1goingdark: AVX2 has no byte-level variable shift: only `vpsllvd`/`vpsllvq` for 32/64-bit lanes, and `vpsllvw` is AVX-512BW-only. The standard AVX2 emulation is multiply-by-power-of-two (mask count to 0..7, `vpshufb` lookup of `2^count`, then `vpmullw`/`vpmulhuw` + mask) — the same fallback LLVM uses until AVX-512BW. That's ~8-10 instructions for 32 bytes, so it may not beat the auto-vectorized path; I'd measure before promising. Happy to try it as a follow-up.

src/u8x32_.rs
bravo1goingdark: `u8x32` is 256 bits — one `ymm`. AVX2's `vpsrlw ymm` already shifts all 32 bytes in one instruction, so a `zmm` would be half-empty. AVX-512BW instructions only help when each lane needs its own count, which is the variable-shift case where we do use them. On `target-cpu=native` (AVX2+AVX-512) the scalar `shr` runs the AVX2 path at 0.8ns.

src/u8x32_.rs
bravo1goingdark: Fair — but no AVX2 instruction narrows 32 `u16` → 32 `u8` in one shot; `vpackuswb` packs 128 bits at a time and scrambles lane order, so the extra extract/unpack/insert steps just fix that. After inlining, LLVM collapses it to a single `vpmovwb` on the AVX-512 paths, which could call `vpmovwb` directly. Happy to simplify.

RagnarGrootKoerkamp:
#319 is going to superseed this.

Also FYI, this PR made right shifts 3x slower than left shifts on my AVX2 laptop, and ~1.5x slower on a AVX-512 machine, so I doubt this ever worked well at all?

bravo1goingdark:
@RagnarGrootKoerkamp do what you gotta do man as long as it benefit the community 👍

PR #315, opened 2026-08-11
Add `from_ne_bytes` and `to_ne_bytes` to all SIMD types

### LINKED ISSUE #144: Add `from_le_bytes` and `to_le_bytes` for SIMD types

bravo1goingdark:
Closes #144

## What

Adds inherent `from_le_bytes` and `to_le_bytes` methods to every SIMD vector type in the crate.

## Why

So you can serialize a vector to bytes (or read one back from bytes) without having to go through `to_array()` and handle each element yourself. It's the SIMD equivalent of `u32::to_le_bytes` / `u32::from_le_bytes`, applied to the whole vector at once.

## How it works

The methods just `transmute` between the vector and a `[u8; size_of::<Self>()]`. Because every SIMD type here is `Pod`, any bit pattern is valid, so:

- `to_le_bytes` returns the raw memory representation of the vector
- `from_le_bytes` is its exact inverse

The "little-endian" name follows the primitive convention, but there's no actual byte-swapping — `to_le_bytes` then `from_le_bytes` always round-trips, regardless of the target's endianness.

## Notes

- Added to the shared `impl_simd!` macro, so all 30 types get it for free
- Tests check the round-trip plus that the bytes match the scalar `T::to_le_bytes` per element, on both endiannesses
- Changelog updated under Unreleased


Lokathor:
I think that if we want to call it `le` bytes then we need to actually enforce that on BE hardware that the swap around happens. Not that there's much BE hardware, but it does exist.

I do like this PR's basic idea, but let's rename all the stuff to being Native Endian (`ne_bytes`), because that's how the standard library names things, so rust programmers are likely to understand that concept already.

bravo1goingdark:
Renamed to `from_ne_bytes`/`to_ne_bytes` per your suggestion — matches the std library naming convention (e.g. `u64::from_ne_bytes`). No BE swap is needed since the functions operate on the raw memory representation of the vector, so they're always inverses regardless of target endianness.

changelog.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+* Added `from_ne_bytes` and `to_ne_bytes` to all SIMD vector types. The byte
```
Noam2Stein: Small nitpick, we probably shouldn't try to explain the entire feature's behavior inside the changelog, since interested users will either go to the new function's docs or check out the PR anyway.

bravo1goingdark: Good call, thanks! Trimmed the changelog entry down to just the summary — the issue link stays for anyone who wants the full context.

Noam2Stein (commented):
Since these functions are simple casts, they can be marked `const`.

Also, a possible use case is to convert lets say `u32x4` to bytes, then perform SIMD operations on those bytes. This suggests the returned type should be `u8xN` instead of `[u8; N]` (I'm not sure that this is the right choice, but lets decide before merging). while for `to_ne_bytes` you can always just `cast::<u32x4, u8x16>`, for future functions `to_le/be_bytes` you cant, then returning `u8xN` is even more useful. To get the array, you could always do `simd.to_ne_bytes().to_array()`. (if we decide to return `u8xN`, this PR becomes blocked on `u8x64`).

Sorry for incrementally asking for changes. I didn't think of these before.

bravo1goingdark:
@Noam2Stein good points!

1. `const` — agreed, both are pure transmutes and the crate already does exactly this in `new`/`to_array`. Added and pushed.

2. `[u8; N]` vs `u8xN` — I'd keep `[u8; N]`:
   - It matches std (`u32::to_ne_bytes` returns `[u8; 4]`). Since the `ne_bytes` naming comes from std, matching the return type keeps semantics predictable for anyone who knows the scalar versions.
   - The names would be misleading otherwise — a function named `to_ne_bytes` returning a vector type feels off.
   - The SIMD-ops-on-bytes use case is already covered today: `cast::<u32x4, u8x16>` does exactly that on all targets. If `to_le/be_bytes` ever land, we can revisit — but I'd rather not block this on adding `u8x64`.

Also, I've rebased onto main — #312 landed after this branch was cut, and the old diff would have reverted the signum/copysign work.

Noam2Stein:
I don't think it'll be inconsistent, since there is a reason for it being different for SIMD vectors. Think of it like `Simd::unbounded_shl` taking an equally sized SIMD type instead of always `u32xN` like `std`, in order to be practical.

`to_*_bytes` does not necessarily imply an *array* of bytes, it can also imply a *SIMD vector* of bytes.

`to_ne_bytes` can be covered by a `bytemuck` cast, but if `to_le/be_bytes` also return a `[u8; _]` they could not, because you need the byte order swapping to happen when the target has an opposite endianness. If we return an array, in order to get a SIMD vector you'll need `u8x16::new(simd.to_le_bytes())` which temporarily unaligns then aligns the value again, which is bad. If we return a SIMD vector, in order to get an array you do `simd.to_le_bytes().to_array()` which *is* optimal.

bravo1goingdark:
Fair point on `unbounded_shl` — wide does deviate from std where SIMD practicality demands it.

Two things keep me on `[u8; N]` for this PR:

1. `u8xN` needs `u8x64` (and its macro partner `i8x64`) for the eight 64-byte types. That's a ~2-type PR by itself; `[u8; size_of::<Self>()]` covers all 30 types today.
2. `u8x16::new(bytes)` is a value-level transmute that stays in registers — no unaligned memory round-trip, same as `bytemuck::cast`.

I agree le/be is where the return type really matters. Happy to land `u8x64` first if that's the direction — @Lokathor, can you weigh in?

Lokathor:
I've got a bit of a headache, so I'm going to write down my train of logic more than usual in case there's some gap I'm missing:

1) We can safely assume that any user of `wide` is also going to use `bytemuck`. It's already in their crate graph, no reason not to.
2) So if you need `u32x4` to `u8x16` then back, we can assume `cast`.
3) If we're assuming that the user has `cast` at their disposal, *why would they ever call these functions*?
4) The only reason I can come up with is because they're used to the `core` methods on primitive ints, and want to use it here.
5) In which case, I think that they would expect the input to be an array, because that's how the core function works.
6) I understand the `u8x16::new(simd.to_le_bytes())` concern, but once the optimizer works on that it won't unalign and re-align the data, it will end up eliminating all the intermediate steps and just transmute, which is either totally free (neon, and x64 int to int) or at most costs you an m128 to m128i movement (x64 float to int). And we don't care about unoptimized code because if we did we wouldn't write in Rust.
7) Since we **are** writing in Rust, we'll take the Rust way out: have a method for both options.
    * We keep `*_ne_bytes` as using `[u8; N]`.
    * We add `*_simd_bytes` for the `u8xN`.

bravo1goingdark:
FYI on where this PR stands with the `*_simd_bytes` plan.

Implementing it means:

1. Adding `u8x64` and `i8x64` as full types (~900 lines) to serve as the `*_simd_bytes` return type for the eight 64-byte vectors.
2. Reworking serde `Deserialize` in the shared macro — serde only supports arrays up to `[T; 32]`, so N=64 wont compile otherwise (custom visitor, behavior-identical).
3. `i8x64::to_bitmask` returning `u64` instead of `u32` — a 64-lane mask doesnt fit in 32 bits.
4. A one-line `ByteSimd` param in each type file — the byte type depends on element size, not lane count, so the macro cant derive it.

My suggestion: keep this PR to `*_ne_bytes` as-is and do `u8x64`/`i8x64` + `*_simd_bytes` as a separate PR — its really its own feature and keeps review focused.

Lokathor:
Point well made, let's keep `*_simd_bytes` as a future possibly.

bravo1goingdark:
@Lokathor this pr is ready for merge whenever you see fit

PR #317, opened 2026-08-14
feat: Dynamic shuffle methods

Noam2Stein:
This PR adds functionality for runtime-index based shuffling, from one or more input vectors.

(Note: This PR copy pastes the `swizzle2` helper functions from the unmerged PR #305)

All functionality is added for all types, not just 8-bit types. My reasoning is that non-8-bit shuffles do have native support on x86, and for other targets, emulating it is quite efficient using a trick with shifts bitors and 8-bit shuffles.

Shuffles from a single input vector are added as the methods:

- `shuffle` (assumes indices are in bounds)
- `zeroing_shuffle` (returns zero for out of bounds indices. this has native support on neon and wasm)
- `wrapping_shuffle` (uses the modulo of indices. this has native support on avx512)

Shuffles from multiple input vectors  (avx512 supports shuffling from 2 vectors, and neon supports from 2, 3 and 4) are added in the `ShuffleExt` trait implemented for arrays of SIMD types, and uses the syntax `[simd_a, simd_b].shuffle(indices)`, with the same three method names.

My reasoning for not just adding `shuffle_2/3/4` methods is that those would result in a large matrix of function names which looks more complicated than this simpler syntax.

# Reasoning behind naming

- "shuffle" is better associated with SIMD, while "swizzle" is commonly associated with shader syntax `vec3.zxy`, so here `shuffle` is more descriptive than `swizzle`

- `swizzle` is already taken by the old function which, depending on the type, uses signed integers for indices or even returns the wrong type (`u8x16::shuffle` returns `i8x16`)

- I think `shuffle` and `shuffle_const` is a better choice than `shuffle_dyn` and `shuffle`. Since dynamic shuffling has multiple variants for out of bounds handling, and constant shuffling does not, its better to put the postfix on constant shuffling, in order to avoid long names like `wrapping_shuffle_dyn`

# Future issue with `shuffle_const`

After this PR, I intend to also add constant-index shuffling via `shuffle_const`. This function will take a const-generic parameter for each lane index (In the future if the type system advances, this can be changed to `shuffle_const<const INDICES: [usize; N]>`, but for now that's impossible).

Because for each different lane-count the function has a different signature, `shuffle_const` cannot be added to `ShuffleExt` directly, since that requires a uniform signature. To support multi-input constant-index shuffles, we would need to add a trait per supported lane count: `Shuffle2Ext`, `Shuffle4Ext`, (same for 8, 16, 32, 64?).

This could be a reason to not go with the array extension trait design, and instead use inherit methods (e.g., `f32x4::shuffle_n([simd_a, simd_b], indices)`?). What should be done with this?

# Deprecated old methods

This PR deprecates `swizzle` since `zeroing_shuffle` has its exact guarantees, and deprecates `swizzle_relaxed` replaced with `shuffle`. The old type-specific guarantees made for `swizzle_relaxed` are retained for `shuffle` for the appropriate types, so old users don't loose any guarantees.

`swizzle_half` and `swizzle_half_relaxed` are not deprecated, since they are not covered by the new methods. I have a design for making those consistent with the new methods, but chose to separate that into another PR.

# Before merging

- Is the array syntax with `ShuffleExt` used here really the best choice? Also consider the issue with `shuffle_const`
- Are there better names for these methods?


Lokathor:
1) the array system is actually a lot better looking than what i would have done with like a `shuffle_2(a, b, indexes)` free function. i mean both the free function form and the array form are "hard to find" if you're using the IDE to look through the method completion options on a value, but something like `x.shuffle2_alongside(y, indexes)` doesn't seem particularly better either. And adding that form later isn't blocked by this array form. So let's consider that part solved.
2) the only way that i can think that would possibly be "better" is to make it `shuffle_zeroing` and `shuffle_wrapping` so that when the user types `x.shuf` and hits Ctrl+Space to method complete, the IDE will prompt them showing 3 options to pick from. Similar to the `sort` and `sort_unstable` naming on slices. Rust should not *require* an IDE to be written, but there's no reason we can't nudge things to the IDE's basic assumptions will still help the user as much as possible.

Noam2Stein:
First lets finish with #318 then ill update this PR.

RagnarGrootKoerkamp:
Just a small note on this: I noticed that the AVX-512 `permutexvar` methods use an 8-bit lane index, and so work mod 256. That's different than both the original relaxed and non-relaxed versions, and with an 8-bit modulo I suspect overflow will be rare in practice, so it would be nice to have a version that just uses that function directly.

(But then again, a mask by `0xff` isn't too bad either and might be optimized away anyway.)

Noam2Stein:
If I understood correctly, this is actually exactly what I was thinking.

Since out of bounds indices are rare, the default function should not handle them. `shuffle` assumes indices are in bounds, and directly uses intrinsics, including neon and wasm zeroing intrinsics, avx512 wrapping intrinsics, and sse/avx mixed behavior intrinsics.

Then you can opt-in into `shuffle_zeroing` and `shuffle_wrapping`, which do handle out of bounds indices.

PR #319, opened 2026-08-14
More efficient shifts by a constant for `u8xL`

RagnarGrootKoerkamp:
~~Followup on #318; only the last commit is new here.~~

This casts to `u16x(L/2)` and does the shift there.
It also computes a mask of all the bits that cross lanes and masks those away.

This is much less code that the previous version, so I assume it'll be faster, but I haven't benchmarked it yet.

Instead of `splat(scalar-computed-mask)` we could also splat the constants and then re-use the shift so that the promotion to `u128` shift argument happens only once; not sure what's better.

It doesn't work for signed types since those do sign extension. But I think it'll probably be faster to do two `u16x(L/2)`-based shifts rather than 'upgrading' to `u16xL` and then 'downgrading' back to `u8xL`.


Noam2Stein:
Since I don't think anybody reviewed this yet, and #318 has been merged, could you rebase this branch on top of main so that the diff is easier to read?

RagnarGrootKoerkamp:
Done. And thanks for merging #318!

src/u8x32_.rs
```diff
@@ -639,11 +641,22 @@ impl_simd_uint! {
...
   #[inline]
   pub fn unbounded_shl_scalar(self, rhs: u32) -> Self {
-    // For x86, this technically can be done explicitly by converting
-    // to `u16` or `u32` then converting back after multiplication, but that
-    // may not actually be faster than auto-vectorization.
-    let [self_a, self_b] = cast::<u8x32, [u8x16; 2]>(self);
-    cast([self_a.unbounded_shl_scalar(rhs), self_b.unbounded_shl_scalar(rhs)])
+    pick! {
+      if #[cfg(target_feature="avx2")] {
+        if rhs >= 8 {
```
Noam2Stein: Is this if else faster than `bitand(result, splat(rhs < 8))`?

If this compiles to a conditional jump, and this function is used for situations where overflow happens a lot, the condition could be hard to predict for the CPU.

Instead I *think* it'll be faster to insert the `rhs` check into the scalar logic that's currently in `crossed` (something like `let mask = u16x16::splat(!((0x00FF_u16 << shift) & 0xFF00) & (rhs < 8));`)

(Edit: I now see that this `if` also exists in previous implementations)

src/u8x16_.rs
```diff
@@ -958,26 +958,11 @@ impl_simd_uint! {
...
-          self_array[12].wrapping_shl(rhs),
-          self_array[13].wrapping_shl(rhs),
-          self_array[14].wrapping_shl(rhs),
-          self_array[15].wrapping_shl(rhs),
-        ])
+        let values: u16x8 = cast(self);
+        let shift = rhs & 7;
+        let shifted = values << shift;
+        let crossed = u16x8::splat((0x00FF_u16 << shift) & 0xFF00);
+        cast(shifted & !crossed)
```
Noam2Stein: This probably gets optimized away anyway, but could it be faster to perform the bitnot before splatting the crossed mask? (this applies to all places with this optimization)

Noam2Stein:
Nice optimization, remember to update the changelog.

src/u8x16_.rs
RagnarGrootKoerkamp: Yeah; this was the same speed but cleaner indeed.

src/u8x32_.rs
RagnarGrootKoerkamp: Indeed, we can avoid the `if`. See main comment.

RagnarGrootKoerkamp:
Allright, it took some time to create proper benchmarks. Indeed the new version is better without explicit `if` condition. Instead, we can use that the [AVX intrinsics](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#=undefined&techs=SSE_ALL,AVX_ALL&text=sll_epi16&ig_expand=6134,6131,6134) already do the `if rhs > 15` check explicitly, and so the result of the **unbounded** shift is 0 for free in these cases. So then the `crossed` mask can just be whatever and we can use a cheap/default wrapping-shift on the scalar `rhs` used in the mask.

Below are the benchmark results on my laptop (intel skylake with AVX2 but not AVX-512) and a server (amd epyc with AVX-512). Rows are different number of lanes and the different operations, while columns are:
- `in-bounds`: the shift is <8 vs `out-of-bounds`: the shift is <16
- `reused`: a single shift for 128 loop iterations vs `random`: a pseudo-random shift per loop iteration
- `seq`: each iteration modifies the previous iterations result vs `par`: each iterations works on a separate value.
There's a bunch of `black_box` sprinkled around which causes some reads/writes from/to L1 which make these timings different from a max-throughput benchmark, but better for a stable comparison.

Also had to add a bunch of `#[inline(always)]` which also needs the unstable `#![feature(stmt_expr_attributes)]` unfortunately.

<details><summary>laptop, before</summary>
<pre>
                      in-bounds                                   out-of-bounds                              
width    shift        reused_seq reused_par random_seq random_par reused_seq reused_par random_seq random_par
u8x16    shl                 2.3        0.8        2.5        2.3        2.3        0.8        2.5        2.3
u8x16    shr                 2.4        0.8        2.5        2.3        2.4        0.9        2.5        2.3
u8x16    unbounded_shl       2.3        0.8        2.5        2.3        1.5        0.8        5.4        5.4
u8x16    unbounded_shr       2.4        0.8        2.5        2.3        1.4        0.8        5.4        5.4

u8x32    shl                 2.4        0.8        2.6        2.3        2.4        0.8        2.6        2.3
u8x32    shr                 5.6        2.4        5.6        2.7        5.5        2.4        5.6        2.7
u8x32    unbounded_shl       2.4        0.8        2.6        2.3        1.5        0.7        5.5        5.4
u8x32    unbounded_shr       5.6        2.4        5.5        2.7        3.0        1.5        5.6        5.6

u8x64    shl                 2.7        1.5        2.8        2.4        2.7        1.5        2.8        2.4
u8x64    shr                 6.2        4.7        6.3        5.0        6.2        4.7        6.3        5.0
u8x64    unbounded_shl       2.7        1.5        2.8        2.4        1.7        1.5        5.7        5.7
u8x64    unbounded_shr       6.2        4.7        6.3        5.0        3.4        3.0        6.1        6.2
</pre>
</details>

<details><summary>laptop, after</summary>
<pre>
                      in-bounds                                   out-of-bounds                              
width    shift        reused_seq reused_par random_seq random_par reused_seq reused_par random_seq random_par
u8x16    shl                 2.3        0.8        2.5        2.3        2.3        0.8        2.5        2.3
u8x16    shr                 2.3        0.8        2.5        2.3        2.3        0.8        2.5        2.3
u8x16    unbounded_shl       2.3        0.8        2.5        2.3        2.3        0.8        2.5        2.3
u8x16    unbounded_shr       2.3        0.8        2.5        2.3        2.3        0.8        2.5        2.3

u8x32    shl                 2.4        0.8        2.7        2.3        2.5        0.8        2.6        2.3
u8x32    shr                 2.4        0.8        2.6        2.3        2.4        0.7        2.6        2.3
u8x32    unbounded_shl       2.4        0.8        2.6        2.3        2.4        0.8        2.6        2.3
u8x32    unbounded_shr       2.5        0.8        2.6        2.3        2.5        0.8        2.6        2.3

u8x64    shl                 2.7        1.4        2.7        2.5        2.7        1.4        2.7        2.6
u8x64    shr                 2.7        1.4        2.7        2.5        2.7        1.4        2.7        2.5
u8x64    unbounded_shl       2.7        1.4        2.7        2.5        2.7        1.4        2.8        2.5
u8x64    unbounded_shr       2.7        1.4        2.8        2.5        2.7        1.4        2.7        2.5
</pre>
</details>


<details><summary>server, before</summary>
<pre>
                      in-bounds                                   out-of-bounds                              
width    shift        reused_seq reused_par random_seq random_par reused_seq reused_par random_seq random_par
u8x16    shl                 4.1        0.5        4.6        1.6        4.1        0.5        4.2        1.6
u8x16    shr                 4.1        0.5        4.6        1.6        4.1        0.5        4.2        1.6
u8x16    unbounded_shl       4.1        0.5        4.6        1.6        2.3        0.5        4.9        4.8
u8x16    unbounded_shr       4.1        0.5        4.6        1.6        2.3        0.5        4.9        4.8

u8x32    shl                 4.9        0.7        5.1        1.6        4.9        0.7        5.1        1.6
u8x32    shr                 5.0        0.9        5.3        1.7        5.0        0.9        5.3        1.7
u8x32    unbounded_shl       4.9        0.7        5.1        1.6        2.7        0.6        5.0        5.0
u8x32    unbounded_shr       5.0        0.9        5.3        1.7        2.7        0.8        5.0        5.0

u8x64    shl                 3.5        1.1        3.5        1.7        3.5        1.1        3.5        1.7
u8x64    shr                 5.2        1.9        5.3        1.9        5.2        1.9        5.3        1.9
u8x64    unbounded_shl       3.5        1.1        3.5        1.7        2.2        1.3        4.9        4.9
u8x64    unbounded_shr       5.2        1.9        5.3        1.9        2.9        1.5        5.0        5.0
</pre>
</details>


<details><summary>server, after</summary>
<pre>
                      in-bounds                                   out-of-bounds                              
width    shift        reused_seq reused_par random_seq random_par reused_seq reused_par random_seq random_par
u8x16    shl                 2.5        0.5        2.9        1.6        2.5        0.5        2.9        1.6
u8x16    shr                 2.5        0.5        3.3        1.9        2.5        0.5        3.3        1.9
u8x16    unbounded_shl       2.5        0.5        2.9        1.6        2.5        0.5        2.9        1.6
u8x16    unbounded_shr       2.5        0.5        3.3        1.9        2.5        0.5        3.3        1.9

u8x32    shl                 2.8        0.5        3.6        1.9        2.8        0.5        3.2        1.9
u8x32    shr                 2.8        0.5        3.6        1.8        2.8        0.5        3.6        1.8
u8x32    unbounded_shl       2.8        0.5        3.6        1.9        2.8        0.5        3.2        1.9
u8x32    unbounded_shr       2.8        0.5        3.6        1.8        2.8        0.5        3.6        1.8

u8x64    shl                 3.2        1.1        3.7        1.9        3.2        1.1        3.7        1.9
u8x64    shr                 3.2        1.1        3.7        2.0        3.2        1.1        3.7        2.0
u8x64    unbounded_shl       3.2        1.1        3.7        1.9        3.2        1.1        3.7        1.9
u8x64    unbounded_shr       3.2        1.1        3.7        2.0        3.2        1.1        3.7        2.0
</pre>
</details>

(sorry for the messed up linebreaks; I can't seem to fix them)

Pretty much everything gets better. The only case that does _not_ get better is the artificial one where we reuse the shift many times in a loop, and we do an `if` statement to see if it's out-of-bounds. In the old case this would lead to a highly predictable branch that just skips most of the work.

Also as mentioned in #313, the implementation added there for `u8x32` for right shifts seems worse than the equivalent old implementation for left shifts.

On my laptop, the perf difference is mostly neutral compared to the pre-#313 baseline, so I guess auto-vectorization worked well already.

On the server, improvements seem to be quite good.

Lokathor (approved):

PR #320, opened 2026-08-15
[DO NOT MERGE] feat: Compile-time-index shuffling

### LINKED ISSUE #194: Feature request: shuffle methods
Looks like LLVM autovectorizer struggles with turning array operations into shuffles, so direct control would be very welcome. This would be particularly useful for things like bitonic sorting.

Noam2Stein:
Resolves #194.

This PR is unfinished and only contains the public API. No implementations yet. I wanted to get approval for the API before spending time on implementations.

Note that this PR is rebased on top of #317, so to review see the last commit.

# Implementations

Implementing constant shuffles *is* possible on stable, using inline assembly on x86 where we pass `const { ... logic that turns indices into mask ... }` as the immediate. Neon only has support for specific permutations. Wasm intrinsics take indices just like we do so we can pass our parameters directly.

The problems mentioned in this PR only concern how the public API looks, not how to implement the functions.

# Future public API

The public API I'd want is: `fn shuffle_const<const INDICES: [usize; N]>(self) -> Self` for vectors, and that same function in `ShuffleExt`. This would be a clean extension of #317. However this is impossible on stable. It requires `min_adt_const_params` and `generic_const_parameter_types`.

# Current API

The solution I implemented here is to define `fn shuffle_consts<const I0: usize, const I1: usize, ...>`, with the hope that the nightly features will be stabilized in the future and this function can be deprecated and replaced.

One downside of this is that function signatures in documentation get extremely long for types with a lot of lanes (I can't screenshot into here, but you can see with `cargo doc`).

For multi-input shuffles, its impossible to put `shuffle_consts` in `ShuffleExt`, since the function has a different signature based on `N`. The solution I implemented here is to define additional traits `Shuffle2Ext`, `Shuffle4Ext` (and same for 8, 16, 32) and put the function there.

A big downside is that this adds multiple traits that complicate the public API, and we may deprecate these traits in the future.

# Questions before implementing the functions

- Is this API a good idea?
- Is there is a better API that avoids the duplicate traits?
- Is it better to give up on these methods until the type system features are stabilized in 10 years? (though until then `std::simd` will replace this lib)


Noam2Stein:
After thinking about it, this API is less problematic than I thought it was, so Ill start implementing and make it into a new PR in order to remove these rebase commits.

PR #325, opened 2026-08-27
feat: mixed SIMD types in `select`

Noam2Stein:
This PR adds implementations for the `Select` trait that allow selecting floats and signed/unsigned integers from each other (e.g., select `f32x4` with `u32x4` mask). Implementations are only for equally sized `T` types and the same `N`.


Noam2Stein:
@Lokathor When you have the time, could you make a release? A lot of improvements have been made since `1.6.1`

Lokathor:
Put out 1.7

